PR #8, opened 2017-04-24
Userprofile json schema

johngian:


johngian:
@andrewkrug @gdestuynder can you take a look?

This is the schema based on: https://github.com/mozilla-iam/ldap-postgres-reference/blob/master/doc/User_profile.json

FYI i am not enforcing any value checks atm. At this point I am only checking the schema and the types.

johngian:
For more information about JSON schema:
http://json-schema.org/

gdestuynder:
This looks alright to me. Note, I found some issues with our current schema for the user profile:

- some attributes cannot be overwritten (auth0 fails if so)
- some attributes are relying upon by RP's and might need to be preserved/legacy, in particular "groups" attribute
- I can set you up with a mocked up profile that uses the above on auth0-dev to get a better idea? I can also update the profile demo in the POC but I believe your schema file will be *the* schema (though we could also have a rendered "demo" for readability. In any case this should leave in the CIS repos, so.. :)

johngian:
@gdestuynder A couple of comments (although I believe we shouldn't be blocked at this point):

* https://github.com/mozilla-iam/cis/pull/8/files#diff-16da26b86d8a5654e680f1d3757537f4R192
  * I added a regex to match the [membership keys ](https://github.com/mozilla-iam/ldap-postgres-reference/blob/master/doc/User_profile.json#L94)
    * Should we go with a regex to allow any arbitrary `membership` name (matching alphanumeric + dot, dash, underscore, space)? I would prefer to explicitly define those although i am not sure if we know beforehand the `membership` keys.

* I am opening a new PR a json factory following this schema, although some real data would be useful too.

gdestuynder:
@johngian I believe we should also whitelist the very exact keys. We do not want new extra membership types unless specifically discussed/approved and thus PR'd
NOTE: as per Dustin there is a typo in the "mozillians.org" example (it spells "mozillian.org" instead, which is wrong of course)

djmitche:
Some descriptions would help.  It's not clear what a "membership" is, and why it has an array of things in it.  And what the difference is between "name" and "roles".

johngian:
@gdestuynder I updated the PR fixing the following issues:

* We now do not validate against auth0 specific attributes
* Memberships check against an arbitrary list of service names

gdestuynder:
@johngian we had additional discussions regarding the compatibility issue without `groups` and the clarity of the authorization scheme.
If you can read through my commit this would be a possible change: https://github.com/gdestuynder/ldap-postgres-reference/blob/group_test/doc/User_profile.json#L82 (I did not PR that as this is just a potential change). We should discuss this during the tech alignment though, and it's probably fine to merge the PR now and amend afterwards as well (unless you're reading this and think my change is awesome and should just be implemented of course, in which case, feel free !! ;-)

johngian:
Let's merge this as it is and change the schema accordingly to avoid blocking the validation development.

andrewkrug:
++ @johngian

PR #9, opened 2017-05-05
Implement AWS KMS envelop encryption

johngian:
* Add setup.py for python packaging
* Add project dependencies with hash-checking
* Implement encrypt/decrypt methods
* Setup app configuration


cis/encryption.py
```diff
@@ -0,0 +1,52 @@
...
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+
+from cis import settings
+
+
+kms = boto3.client('kms')
+
+
+def encrypt(plaintext):
+    """Encrypt CIS payload using keys derived from KMS."""
```
akatsoulas: What about adding a couple of more lines in each docstring (enrcypt/decrypt) explaining the args and the general behavior of these functions (although the later is rather self-evident).

requirements.txt
```diff
@@ -18,3 +18,77 @@ python-dateutil==2.6.0 \
 s3transfer==0.1.10 \
     --hash=sha256:bc52f38637f37572ae180c08a2dd87a4825b46713797633aab6e1a42d6b2b8ff \
     --hash=sha256:ba1a9104939b7c0331dc4dd234d79afeed8b66edce77bbeeecd4f56de74a0fc1
+python-decouple==3.0 \
```
johngian: @akatsoulas @andrewkrug I am not sure what's the proper way to proceed with the project dependencies

* ``setup.py`` install requires are better to package cis
* ``requirements.txt`` supports hash checking to make sure that allow more fine grained integrity across envs (also improves deployment security by making the env reproducible).

thoughts ?

johngian:
I also added a couple of tests for encryption/decryption.

johngian:
@akatsoulas @andrewkrug r?

cis/encryption.py
```diff
@@ -0,0 +1,63 @@
...
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+
+from cis import settings
+
+
+kms = boto3.client('kms')
+
+
+def encrypt(plaintext):
```
andrewkrug: May want to consider adding encryption context as optional arg.  In case we want to scope a specific set of attributes to specific grants.

andrewkrug: May want to consider adding encryption context as optional arg.  In case we want to scope a specific set of attributes to specific grants.

cis/encryption.py
```diff
@@ -0,0 +1,63 @@
...
+
+    return {
+        'ciphertext': ciphertext,
+        'ciphertext_key': ciphertext_key,
+        'iv': iv,
+        'tag': encryptor.tag
+    }
+
+
+def decrypt(ciphertext, ciphertext_key, iv, tag):
```
andrewkrug: If context is added to encrypt it should also be in decrypt.

tests/test_encryption.py
```diff
@@ -0,0 +1,56 @@
...
+            'ciphertext': base64.b64decode(self.test_artifacts['expected_ciphertext']),
+            'ciphertext_key': test_kms_data['CiphertextBlob'],
+            'iv': test_iv,
+            'tag': base64.b64decode(self.test_artifacts['expected_tag'])
+        }
+
+        from cis.encryption import encrypt
+        self.assertEqual(expected_result, encrypt(b'foobar'))
+
+    @patch('cis.encryption.kms')
```
andrewkrug: No comment.  But this is a cool way to test this without recording API calls.  <3

PR #13, opened 2017-05-10
Update travis and setup confs.

akatsoulas:


setup.cfg
```diff
@@ -1,2 +1,3 @@
 [flake8]
 max-line-length=99
+exclude=dockerfiles
```
johngian: ``flake8`` doesn't check dockerfiles anyway

.travis.yml
```diff
@@ -2,12 +2,11 @@ language: python
 python:
   - "3.6"
 before_install:
-  - "pip install flake8"
-  - "flake8 cis"
-  - "flake8 tests"
+  - pip install flake8
```
johngian: This is exactly the same thing in .yaml

akatsoulas: Agreed, I am just in favor of simplicity. The whole pr is nit.

.travis.yml
```diff
@@ -2,12 +2,11 @@ language: python
 python:
   - "3.6"
```
johngian: If you want to avoid using ``"`` then replace it across all the file.

PR #14, opened 2017-05-11
CIS data validation

johngian:


cis/validation.py
```diff
@@ -0,0 +1,39 @@
...
+
+from jsonschema import validate as jsonschema_validate
+
+from cis.encryption import decrypt
+
+
+CIS_SCHEMA_JSON = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'schema.json')
+
+
+with open(CIS_SCHEMA_JSON, 'r') as schema_data:
```
akatsoulas: Please check that the file exists.

cis/validation.py
```diff
@@ -0,0 +1,39 @@
...
+        decrypted_payload = decrypt(**payload)
+    except:
+        return False
+
+    try:
+        # Encode binary decrypted payload to unicode string
+        # Load json as dictionary object
+        # Check for content validity against jsonschema
+        json_object = json.loads(decrypted_payload.decode('utf-8'))
+        jsonschema_validate(json_object, CIS_SCHEMA)
```
akatsoulas: Probably you need to make sure that the CIS_SCHEMA exists

cis/validation.py
```diff
@@ -0,0 +1,39 @@
...
+    Validates the payload passed to CIS.
+
+    :payload: Encrypted payload based on the output of `cis.encryption.encrypt` method
+    """
+
+    try:
+        # Decrypt payload coming from CIS using KMS key
+        # This ensures that publisher is trusted by CIS
+        decrypted_payload = decrypt(**payload)
+    except:
```
akatsoulas: Can we catch specific exceptions here so that we can avoiding silencing whatever may cause them? (I don't have a good idea atm)

cis/validation.py
johngian: File always exists. Its in the codebase and its going to be shipped with the package.

cis/validation.py
johngian: Not really, I already though about it. The reasons are:

* The exceptions are like 20 (considering that its both AWS KMS API exceptions and AES cryptography exceptions)
* In all the cases we need to return ``False``.

What's missing but its outside of the scope is:

* Handle logs to cloudwatch or another log provider
* Pass it back to the stream (although this needs discussion and its completely outside the current OKR goal)

cis/validation.py
johngian: ``CIS_SCHEMA`` always exists.
https://github.com/mozilla-iam/cis/blob/master/cis/schema.json

cis/validation.py
akatsoulas: maybe limit to `except Exception as e` then?

johngian:
@akatsoulas PR updated with exception logging

PR #28, opened 2017-05-15
Add method to store CIS data to DynamoDB.

johngian:


cis/validation.py
```diff
@@ -42,3 +44,24 @@ def validate(**payload):
         return False
 
     return True
+
+
+def store_to_vault(data):
```
akatsoulas: Shall we add a check for the data format somewhere in this function?

tests/test_validation.py
```diff
@@ -87,3 +88,63 @@ def test_invalid_kms_key(self, mock_kms):
...
+
+        dynamodb_mock.Table.return_value = dynamodb_table
+        mock_boto.resource.return_value = dynamodb_mock
+
+        data = {
+            'foo': 'bar',
+            'foobar': 42
+        }
+
+        from cis.validation import store_to_vault
```
akatsoulas: Is this causing circular dependencies? If not, let's move it to the top.

tests/test_validation.py
```diff
@@ -87,3 +88,63 @@ def test_invalid_kms_key(self, mock_kms):
...
+        dynamodb_table.put_item.side_effect = Exception('DynamoDB exception')
+        dynamodb_mock.Table.return_value = dynamodb_table
+        mock_boto.resource.return_value = dynamodb_mock
+
+        data = {
+            'foo': 'bar',
+            'foobar': 42
+        }
+
+        from cis.validation import store_to_vault
```
akatsoulas: Same here.

tests/test_validation.py
```diff
@@ -87,3 +88,63 @@ def test_invalid_kms_key(self, mock_kms):
             from cis.validation import validate
             is_valid_payload = validate(**payload)
         self.assertFalse(is_valid_payload)
+
+
+class DatabaseTest(unittest.TestCase):
```
akatsoulas: Shall we add another test with malformed data?

cis/validation.py
johngian: No let's keep it atomic. The validation is happening in a different level.

tests/test_validation.py
johngian: Validation is tested in the TestCase above.

tests/test_validation.py
johngian: We need this here for setUp to work correctly (we are patching env vars) in all our testcases

tests/test_validation.py
johngian: Same as above.

PR #32, opened 2017-05-22
Add contributing file from https://github.com/mozilla-iam/mozilla-iam

gdestuynder:


gdestuynder:
Badges are a great idea and i opened https://github.com/mozilla-iam/mozilla-iam/issues/26 for now (just because ive never done it - id be happy if we "just did it" though)

PR #39, opened 2017-05-25
fallback to environment if decouple config not present

andrewkrug:
This adds a default fall through to the decouple config initialize to support using lambda environment vars instead of settings.ini if necessary.


andrewkrug:
Not needed.

Issue #16, opened 2017-05-12
[TESTING] Create a mocked profile input script


andrewkrug:
We'll need a way to simulate change events to the CIS library.  This task captures the following:

- [x] Generates a mocked user profile
- [x] Inputs to CIS first lambda function
- [x] Validates success / failure ( should have both know fail and known success test cases )

^^ Arguably the last task in this card is an end to end test with a greater set of faked profiles.  @tyearns suggest marking this one done and creating a new task for "end to end test with a number of user profiles"

andrewkrug:
@johngian this seems like as good a place to start as any... mocking event.json stuff to the input validator.  Should we add a fake_profile function to the CIS core library.

andrewkrug:
nemo is working on this story right now so I'm assigning and linking the method that he will attempt in an effort to keep the card up to date. https://github.com/json-schema-faker/json-schema-faker

johngian:
PR opened here:
#37

PR #43, opened 2017-05-25
Add README on how to run data faker.

johngian:


johngian:
This is relevant to #16

PR #45, opened 2017-05-26
Add function to validate a json string against schema.

johngian:


andrewkrug:
@johngian Does this change the way I need to call the validator for an encrypted payload?

andrewkrug:
Answered my own question... looks good to me.

PR #57, opened 2017-06-08
Add helper method to invoke lambda validator function.

johngian:


johngian:
@andrewkrug r?

I haven't really used boto3 lambda invoke. What do you think about the method?

cis/settings.py
```diff
@@ -4,3 +4,4 @@
 ARN_MASTER_KEY = config('CIS_ARN_MASTER_KEY')
 DYNAMODB_TABLE = config('CIS_DYNAMODB_TABLE')
 KINESIS_STREAM_ARN = config('CIS_KINESIS_STREAM_ARN')
+LAMBDA_VALIDATOR_ARN = config('CIS_LAMBDA_VALIDATOR_ARN')
```
andrewkrug: Can we make this an optional config parameter since it's not always needed.  In other words always fall back to None?

cis/streams.py
```diff
@@ -26,10 +26,41 @@ def publish_to_cis(data, partition_key):
...
         StreamName=KINESIS_STREAM_ARN.split('/')[1],
-        Data=json_payload,
+        Data=payload,
         PartitionKey=partition_key
     )
 
     return response
+
+
+def invoke_cis_lambda(data):
```
andrewkrug: If the LAMBDA VALIDATOR ARN is optional can we also make this raise if the function is called and the lambda can not be found instead of relying on boto3

cis/streams.py
```diff
@@ -26,10 +26,41 @@ def publish_to_cis(data, partition_key):
...
+    Invoke lambda function in front of the CIS pipeline with data to be pushed to CIS
+
+    :data: Data to be published to CIS (dict)
+    """
+
+    payload = prepare_payload(data)
+
+    response = lambda_client.invoke(
+        FunctionName=LAMBDA_VALIDATOR_ARN,
+        InvocationType='RequestResponse',
```
andrewkrug: This is the correct type of invocation.  This will block on return until validator sends the jsonified response.

johngian:
@andrewkrug regarding the optional config parameters i was thinking to make them all lazy. That means that it is going to be required only when called.

johngian:
Thoughts?

andrewkrug:
@johngian lazy evals sound good to me.  I know for most of it right now if the ENV var isn't present the function blows up.  

Mostly I'm trying to avoid having more config data in each lambda than is needed.  So as long as we achieve that I don't have a strong opinion on how.

cis/settings.py
johngian: I refactored the way we handle config and everything is optional.

cis/streams.py
johngian: Fixed

johngian:
@andrewkrug PR updated

andrewkrug:
@johngian reviewed.

PR #60, opened 2017-07-17
add utils from cis_functions into core

### LINKED ISSUE #55: Remove duplicate code
Having duplicate code is really bad:

* Error prone since you need to apply changes to all the files one by one 
* Its a maintenance hell because you need to keep track of the same copy in multiple different paths

Let's deduplicate:

* utils.py
* clean.sh
* requirements.txt

andrewkrug:
Closes #55 in part 1 of 2

Adding this to cis is step #1.  

Step 2 is to remove the utils.py from cis_functions.

### DIFF
--- a/cis/__init__.py
+++ b/cis/__init__.py
@@ -3,4 +3,5 @@ __all__ = [
     'settings',
     'streams',
     'validation',
+    'utils'
 ]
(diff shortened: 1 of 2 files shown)


andrewkrug:
https://github.com/mozilla-iam/cis_functions/pull/14

PR #64, opened 2017-07-20
Multiple fixes

johngian:
* Add missing __init__ file
* Fix import order


johngian:
Minor fixes (well ``__init__`` is not that minor :smile:) that I found while skimming through the last commits.

johngian:
@akatsoulas r?

PR #70, opened 2017-08-01
Mzlians plugin validation support

gdestuynder:
+flake8

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,3 @@
 [flake8]
-max-line-length=99
+max-line-length=119
+ignore=E266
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -1,2 +1,3 @@
 [flake8]
-max-line-length=99
+max-line-length=119
```
andrewkrug: I think we set this at 120 in all the other projects?

gdestuynder: usually its set at -1 but i dont mind setting 120. pep8 is 100 max btw

PR #71, opened 2017-08-03
oo refactor

andrewkrug:
This refactor of the CIS library implements basic nullObject pattern to improve test-ability of code and lays the foundation for signatures.  

Signatures are not currently enforced all the way through but all objects support them. 

Here's a code sample of publishing to the new stream using the new publisher object. 

This PR also introduces a simplified processor object that collapses the logic in cis_functions. 

```
    event_file = os.path.join(
        os.path.abspath(os.path.dirname(__file__)), 'event.json'
    )

    with open(event_file, 'r') as event_data:
        profile_data = (event_data.read().encode('utf-8'))


    from cis.libs import encryption

    encryptor = encryption.Operation(boto_session=boto3.Session(region_name='us-west-2'))

    event = {
        'publisher': {'id': 'mozillians.org'},
        'signature': {},
        'profile': encryptor.encrypt(profile_data)
    }

    print(event)

    res = handle(event, None)
```


andrewkrug:
@gdestuynder @johngian didn't have time to pep8 all the things.  I can make another pep8 pass post sprint review.

cis/__init__.py
```diff
@@ -1,7 +1,3 @@
 __all__ = [
```
johngian: Either remove or fix this entry. This way when importing everything under module ``cis`` imports nothing.

cis/libs/streams.py
```diff
@@ -0,0 +1,57 @@
...
+        base64_payload = dict()
+        for key in ['ciphertext', 'ciphertext_key', 'iv', 'tag']:
+            base64_payload[key] = base64.b64encode(binary_payload[key]).decode('utf-8')
+
+        base64_payload['signature'] = event.get('signature')
+
+        json_payload = json.dumps(base64_payload).encode('utf-8')
+        return json_payload
+
+
```
gdestuynder: lotsa straight LF ;-)

cis/libs/connection.py
```diff
@@ -0,0 +1,52 @@
...
+        self.profile = profile
+
+        try:
+            # Attempt to init default boto session and raise if fails
+            boto3.setup_default_session(profile_name=self.profile)
+        except Exception as e:
+            raise(e)
+
+    def connect(self):
+        if self.connection_type is None:
```
johngian: ``connection_type`` can only be ``None`` if user selects ``None`` as an argument. By default it is a required argument in the ``__init__`` signature and if it's not supplied python will raise an error anyway.

cis/libs/validation.py
```diff
@@ -0,0 +1,67 @@
...
+        with self.plugin_source:
+            for plugin in self.plugin_load:
+                cur_plugin = self.plugin_source.load_plugin(plugin)
+                try:
+                    if cur_plugin.run(self.publisher, self.user, self.profile_data) is False:
+                        return False
+                    else:
+                        pass
+                except Exception as e:
+                    print(e)
```
gdestuynder: s/print/traceback.print_exc/ ? (or stray debug code?)

cis/libs/connection.py
```diff
@@ -0,0 +1,52 @@
...
+            raise AttributeError(
+                "Could not determine connect type.  Set client or resource."
+            )
+        elif self.connection_type == "client":
```
johngian: Instead of having all these ``if/elif/else`` cases and check against strings I suggest we have a default connection_type that works in most cases and override it with a ``boto`` client object instance/resource/session.

cis/plugins/validation/mozilliansorg_publisher_plugin.py
```diff
@@ -36,8 +36,12 @@ def run(publisher, user, profile_json):
         'groups'
     ]
 
-    # Check for any non-whitelisted attribute modification
-    # Note that no extra attribute may be added at this stage as we already performed schema validation
+    # Check the easiest case.  None type.
```
gdestuynder: this means we need to preimport users, until we have publishers which are allowed to create new users (workday/auth0 i believe)

cis/libs/connection.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        self.service = service
+        self.region = region
+        self.connection_type = connection_type
+        self.profile = profile
+
+        try:
+            # Attempt to init default boto session and raise if fails
+            boto3.setup_default_session(profile_name=self.profile)
+        except Exception as e:
```
johngian: Catch a specific exception here. This way you are catching  and re-raising all possible exceptions which is equivalent of not having this try/except at all.

cis/processor.py
```diff
@@ -0,0 +1,127 @@
...
+                    signature=self.signature,
+                    encrypted_profile_data=self.encrypted_profile_data
+                ).to_kinesis()
+
+                if res['ResponseMetadata']['HTTPStatusCode'] == 200:
+                    res = True
+        elif self.stage == 'streamtoidv':
+            res = self._vault_stage()
+        elif self.stage == 'idvtoauth0':
+            # TBD fold in the authzero logic from idvtoauth0 in cis_functions
```
gdestuynder: nit: s/TBD/FIXME. i usually just go with smth like 'XXX FIXME TBD INSERTURSHERE' so that everyone's favorite IDE picks these up

gdestuynder: would also add that to the libs/authzero.py placeholder

cis/libs/encryption.py
```diff
@@ -0,0 +1,98 @@
...
+        ).decryptor()
+
+        return decryptor.update(ciphertext) + decryptor.finalize()
+
+    def _get_initialization_vector(self):
+        return os.urandom(12)
+
+    def _get_data_key_from_kms(self, encryption_context={}):
+        """
+
```
johngian: Please remove empty line here.

cis/libs/encryption.py
```diff
@@ -0,0 +1,98 @@
...
+class Operation(object):
+    def __init__(self, boto_session):
+        self.config = get_config()
+        self.data_key = None
+        self.iv = None
+
+        try:
+            self.kms = boto_session.client(service_name='kms')
+        except Exception as e:
```
johngian: Same here. Be more specific on the exceptions you catch.

cis/processor.py
```diff
@@ -0,0 +1,127 @@
...
+            pass
+        else:
+            # Unhandled pass for anything not handled in block.  Basically yield to block.
```
gdestuynder: logger.exception('unhandled processor stage type {}'.format(self.stage)) or similar?

cis/libs/encryption.py
```diff
@@ -0,0 +1,98 @@
...
+            self.kms = None
```
johngian: What's the point of having ``kms`` None if you are going to use ``None`` later. ``self.kms`` should never be none. And if ``boto_session.client()`` fails it should raise an error.

cis/processor.py
```diff
@@ -0,0 +1,127 @@
...
+        self.stage = self._get_stage()
+        self.user = None
+
+    def run(self):
+        # Determine what stage of processing we are in and call the corresponding functions.
+
+        self.decrytped_profile = json.loads(self._decrypt_profile_packet())
+
+        if self.stage == 'validator':
+            logger.info('Validator logic activated.')
```
gdestuynder: nit: seems like logger.debug to me, or actually, move that one 1 line up and:
logger.info('Attempting to load stage processor logic: {}'.format(self.stage)) instead

cis/libs/streams.py
```diff
@@ -0,0 +1,57 @@
+import base64
+import json
+from cis.settings import get_config
```
johngian: Please add a line to seperate python/3rd party/local imports.

cis/publisher.py
```diff
@@ -0,0 +1,136 @@
...
+
+        return response
+
+class Change(ChangeDelegate):
+    """Guaranteed change object."""
+    def __init__(self, publisher=None, signature=None, profile_data=None):
+        try:
+            ChangeDelegate.__init__(self, publisher, signature, profile_data)
+        except Exception as e:
+            print(e)
```
gdestuynder: s/trackeback.print_exc/

cis/libs/streams.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def to_kinesis(self):
+        """
+        Publish data to CIS kinesis stream given a partition key.
+    
+        :data: Data to be published to kinesis (dict)
+        :partition_key: Kinesis partition key used to publish data to
+        """
+
+        if self.kinesis_client is None:
```
johngian: The way that you have coded it ``self.kinesis_client`` is always ``None`` when running this method. Please move the initialization to ``__init__``.

cis/libs/streams.py
```diff
@@ -0,0 +1,57 @@
...
+            self.kinesis_client = self.boto_session.client('kinesis')
+
+        event = {
+            'publisher': self.publisher,
+            'profile': self.encrypted_profile_data,
+            'signature': self.signature
+        }
+
+        stream_arn = self.config('kinesis_stream_arn', namespace='cis')
+        stream_name = stream_arn.split('/')[1]
```
johngian: Since we are refactoring this, let's avoid ``split`` on something that we define ourselves. Let's prepare the correct format in the config level.

setup.py
```diff
@@ -1,11 +1,25 @@
...
+    return test_suite
+
+class install(_install):
+    def run(self):
+        self.run_command('build')
+        _install.run(self)
 
 setup(
     name='mozilla-iam-cis',
     version='0.1dev',
```
gdestuynder: probably s/0.1dev/1.0/

cis/libs/streams.py
```diff
@@ -0,0 +1,57 @@
...
+
+        base64_payload = dict()
+        for key in ['ciphertext', 'ciphertext_key', 'iv', 'tag']:
+            base64_payload[key] = base64.b64encode(binary_payload[key]).decode('utf-8')
+
+        base64_payload['signature'] = event.get('signature')
+
+        json_payload = json.dumps(base64_payload).encode('utf-8')
+        return json_payload
+
```
johngian: Please remove trailing empty lines.

cis/libs/validation.py
```diff
@@ -0,0 +1,67 @@
+import os
+import logging
+from pluginbase import PluginBase
```
johngian: Please add a line to separate python dependencies from 3rd party.

tests/old/1_test_encryption.py
```diff
@@ -0,0 +1,81 @@
...
+
+        from cis.libs.encryption import encrypt_payload
+        self.assertEqual(expected_result, encrypt_payload(None, b'foobar'))
+        mock_kms.generate_data_key.assert_called_once_with(
+            KeyId=self.test_artifacts['dummy_kms_arn'],
+            KeySpec='AES_256',
+            EncryptionContext={}
+        )
+
+"""
```
gdestuynder: not currently working? (test_decrypt_payload)

cis/libs/validation.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(self, publisher, profile_data, user=None):
+        self.publisher = publisher
+        self.profile_data = profile_data
+        self.user = user
+
+    def is_valid(self):
+        """Source of truth for all validation options."""
+        s = Schema(self.publisher, self.profile_data, self.user)
+
+        if s.validate() is True:
```
johngian: This is equivalent of ``return s.validate()``.

cis/libs/validation.py
```diff
@@ -0,0 +1,67 @@
...
+            return True
+        else:
+            return False
+
+class Schema(object):
```
johngian: Please move this before ``class Operation`` to improve readability.

cis/libs/validation.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def _initialize_plugin_source(self):
+        plugin_base = PluginBase(package='cis.plugins.validation')
+        plugin_source = plugin_base.make_plugin_source(
+                searchpath=[
+                    os.path.join(
+                        os.path.abspath(
+                            os.path.dirname(__file__)
+                        ),
+                        '../plugins/validation/'
```
johngian: Let's avoid relative paths. I suggest we add a config entry the returns an absolute path. Because:

cis/plugins/validation/json_schema_plugin.py
```diff
@@ -7,12 +7,13 @@
...
+cis_schema_json = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'schema.json')
 with open(cis_schema_json, 'r') as schema_data:
     cis_schema = json.load(schema_data)
 
 
 def run(publisher, user, profile_json):
     # Check for content validity against jsonschema
     jsonschema_validate(profile_json, cis_schema)
-    return True
+
```
johngian: Please remove these new empty lines that you added.

cis/processor.py
```diff
@@ -0,0 +1,127 @@
+"""Iterative Stream Processor plugin for handling steps during data storage."""
+import json
+import logging
+import os
+
+from cis import user
+
```
johngian: Remove this line.

cis/processor.py
```diff
@@ -0,0 +1,127 @@
...
+        logger.info('The result of the change was {r}'.format(r=res))
+        return res
+
+    def _decrypt_profile_packet(self):
+        return self.decryptor.decrypt(
+            ciphertext=self.encrypted_profile_data.get('ciphertext'),
+            ciphertext_key=self.encrypted_profile_data.get('ciphertext_key'),
+            iv=self.encrypted_profile_data.get('iv'),
+            tag=self.encrypted_profile_data.get('tag')
+
```
johngian: Remove empty line.

cis/processor.py
```diff
@@ -0,0 +1,127 @@
...
+        self.kinesis_client = None
+        self.publisher = publisher
+        self.signature = signature
+        self.stage = self._get_stage()
+        self.user = None
+
+    def run(self):
+        # Determine what stage of processing we are in and call the corresponding functions.
+
+        self.decrytped_profile = json.loads(self._decrypt_profile_packet())
```
johngian: Please fix typo and its references.

cis/processor.py
```diff
@@ -0,0 +1,127 @@
...
+
+        if self.stage == 'validator':
```
johngian: Again I would avoid checking against strings and have all these nested ``if/elif/else`` cases which are very error prone and hard to follow. Since you are going the OO way it would be far more elegant to subclass the different stages.

cis/processor.py
```diff
@@ -0,0 +1,127 @@
...
+            ciphertext=self.encrypted_profile_data.get('ciphertext'),
+            ciphertext_key=self.encrypted_profile_data.get('ciphertext_key'),
+            iv=self.encrypted_profile_data.get('iv'),
+            tag=self.encrypted_profile_data.get('tag')
+
+        )
+
+    def _get_stage(self):
+        # Let the object know what phase of operation we are running in.
+        stage = os.environ.get('APEX_FUNCTION_NAME', None)
```
johngian: Let's avoid having cross dependencies of ``cis_functions`` and ``cis``. CIS core is the logic that runs behind the functions. We want to keep them clean and atomic.

By adding this logic here we import code from ``cis`` to ``cis_functions`` that depends on ``cis_functions`` logic to run.

cis/publisher.py
```diff
@@ -0,0 +1,136 @@
+"""First class object for publishing messages to Mozilla Change Integration Service."""
+import base64
+import json
+
+from cis.libs import connection
```
johngian: Please sort imports alphabetically.

cis/publisher.py
```diff
@@ -0,0 +1,136 @@
...
+from cis.libs import encryption
+from cis.libs import validation
+
```
johngian: Remove empty line.

cis/publisher.py
johngian: Let's avoid printing, that's the reason we use ``logger`` instead.

cis/publisher.py
```diff
@@ -0,0 +1,136 @@
...
+        return response
+
+class Change(ChangeDelegate):
+    """Guaranteed change object."""
+    def __init__(self, publisher=None, signature=None, profile_data=None):
+        try:
+            ChangeDelegate.__init__(self, publisher, signature, profile_data)
+        except Exception as e:
+            print(e)
+            print("returning null object")
```
johngian: Same here.

cis/publisher.py
```diff
@@ -0,0 +1,136 @@
...
+        function_name = self.config('lambda_validator_arn', namespace='cis')
+        response = self.lambda_client.invoke(
+            FunctionName=function_name,
+            InvocationType='RequestResponse',
+            Payload=event
+        )
+
+        return response
+
+class Change(ChangeDelegate):
```
johngian: This whole class can be truncated to ``ChangeDelegate``. You are just overriding ``__init__`` which if it fails you log something. This adds much more complexity without any added value.

cis/user.py
```diff
@@ -0,0 +1,66 @@
+"""First class object to represent a user and data about that user."""
+import logging
+from cis.settings import get_config
```
johngian: Seperate imports.

cis/user.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class Profile(object):
+    def __init__(self, boto_session=None, profile_data=None):
+        """
+
```
johngian: Please remove empty line.

cis/user.py
```diff
@@ -0,0 +1,66 @@
...
+    @property
+    def exists(self):
+        if self._retrieve_from_vault() is not None:
+            return True
+        else:
+            return False
+
+    def retrieve_from_vault(self):
+
```
johngian: Same here.

cis/user.py
```diff
@@ -0,0 +1,66 @@
...
+        if self.dynamodb_table is None:
+            self._connect_dynamo_db()
+
+        user_key = {'user_id': self.profile_data.get('user_id')}
+
+        try:
+            response = self.dynamodb_table.get_item(Key=user_key)
+            self.profile_data = response
+        except Exception:
```
johngian: Again please be specific in the ``Exception``.

johngian: This way you are hiding the actual error so we don't know and cannot debug what went wrong.

cis/user.py
```diff
@@ -0,0 +1,66 @@
...
+            logger.exception('DynamoDB GET failed')
+            return None
+
+        return response
+
+    def store_in_vault(self):
+        # Put data to DynamoDB
```
johngian: Use docstrings instead:
```Put data to DynamoDB``.

File: cis/user.py [hunk 40]
```diff
@@ -0,0 +1,66 @@
...
+        try:
+            response = self.dyanmodb_table.put_item(
+                Item=self.profile_data
+            )
+        except Exception:
```
johngian: Same here.

requirements-dev.txt
```diff
@@ -0,0 +1,25 @@
+appdirs==1.4.3
```
johngian: Why are you renaming this file to ``requirements-dev.txt``?

setup.py
```diff
@@ -1,11 +1,25 @@
-from distutils.core import setup
+from distutils.command.build import build
+from setuptools import setup
+from setuptools.command.install import install as _install
 
+import unittest
```
johngian: Please move this to the start of the file. Import order is:

* Complete module imports
* submodule imports

setup.py
```diff
@@ -1,11 +1,25 @@
...
+
+def test_suite():
+    test_loader = unittest.TestLoader()
+    test_suite = test_loader.discover('tests', pattern='test_*.py')
+    return test_suite
+
+class install(_install):
```
johngian: Let's avoid monkeypatching setup utils. This is an antipattern.

tests/old/1_test_encryption.py
```diff
@@ -0,0 +1,81 @@
+import base64
```
johngian: There should be no old and new tests. Please fix naming. Also test names should start with ``test__`` to enable discovery.

tests/old/1_test_streams.py
```diff
@@ -0,0 +1,95 @@
+import json
```
johngian: Same here.

tests/old/1_test_validation.py
```diff
@@ -0,0 +1,176 @@
+import base64
```
johngian: Same here.

johngian:
Please fix all PEP8 issues.

johngian:
Please squash all commits to one or more atomic commits.

cis/libs/validation.py
gdestuynder: @johngian seems your comment was cut

andrewkrug:
@johngian @gdestuynder I fixed most of the asks here.  Some of the things I left in place, like the superclassing of the parent objects because they make the cis_functions more testable.  I left the stage environment variable in the code base as well due to the fact that you can either : a. set the environment var or just not import the processor code.  All processors will be apex functions.  This does not effect the ability of non-apex things to publish events.  

Overall I've also cleaned up exception handling and print statements and replaced with loggers.  I'm sure I've missed something though.

cis/__init__.py
andrewkrug: fixed

cis/libs/connection.py
andrewkrug: This was intentional to support a desktop use case.  But... I've eliminated the block now so it will just blow up if no boto profile exists.

cis/libs/connection.py
andrewkrug: This was also intentional to support a test case.  However, this has now been removed as well.

cis/libs/connection.py
andrewkrug: I'm only currently passing around boto3 session objects.  So I've left the code as is but made it default to session as the default type.

cis/libs/encryption.py
andrewkrug: This is a fall through to avoid having to patch / mock KMS.  I've left the code as is but improved the exception handler.

cis/libs/encryption.py
andrewkrug: Done

cis/libs/streams.py
andrewkrug: Done

cis/libs/streams.py
andrewkrug: It's initialized opportunistically if not passed in with the object constructor.  This makes it easy to use moto in the test pattern.

cis/libs/streams.py
andrewkrug: This is an enhancement I'd rather grab after merge.

cis/libs/streams.py
andrewkrug: Done

cis/libs/streams.py
andrewkrug: Done

cis/libs/validation.py
andrewkrug: Done

cis/libs/validation.py
andrewkrug: I elected to drop in loggers here. I'm all for improving the bubble up to CloudWatch so let's decide on the best way and make all the exception logging support that.

cis/libs/validation.py
andrewkrug: Let's file a bug for enhancement on this issue.

cis/plugins/validation/json_schema_plugin.py
andrewkrug: Done

cis/plugins/validation/mozilliansorg_publisher_plugin.py
andrewkrug: We need to talk about this a bit.

cis/processor.py
andrewkrug: Done

cis/processor.py
andrewkrug: Unsure what the typo is here... Left as is.

cis/processor.py
andrewkrug: This is an enhancement.  I know this block of code is a little unwieldy and smells of non-oo design pattern. I could collapse this into sub-classed objects now.  But it works now... So let's merge and then enhance it together until it's what we want.

cis/processor.py
andrewkrug: Same as above.... let's agree on structured log format and then make another paired pass through the code base until we're all happy with the logging.

cis/processor.py
andrewkrug: didn't do this yet.

cis/processor.py
andrewkrug: Agree... but let's get this in a logger pass.

cis/processor.py
andrewkrug: Done

cis/processor.py
andrewkrug: I'm going to leave this here for now.  But I agree with you at a functional level.  The two should be independent of one another. Let's reach an agreement on how CIS should identify whether it's running in a stream processor in lambda or not.  

FYI I don't see a future where we have stream processors that don't run in lambda.

cis/publisher.py
andrewkrug: Done

cis/publisher.py
andrewkrug: Done

cis/publisher.py
andrewkrug: This is a common pattern for handling testing where you need a guaranteed object out the other side.  I can see us adding more logic to this.  So I'm inclined to leave it in place.

cis/user.py
andrewkrug: Done

cis/user.py
andrewkrug: Done

cis/user.py
andrewkrug: Done

cis/user.py
andrewkrug: Ironically I ported this from what was in master branch.  But I've removed these comments completely now.  I think the method names are explicit enough such that they don't need a comment.

requirements-dev.txt
andrewkrug: In order to include testing-only depends like moto.

setup.py
andrewkrug: Removed.

setup.py
andrewkrug: unsure how we track versions of this.

tests/old/1_test_encryption.py
andrewkrug: Removed old tests.

tests/old/1_test_encryption.py
andrewkrug: Fixed.

andrewkrug:
Responded to almost every review request.  I'd like to suggest that we do the following:

A. Accept this as is and then merge the corresponding CIS functions.
B. Reject this pull completely and then pull into a development branch where we all iterate.
C. Reject the attempt completely and go back to the code base the way it was, then spend another full cycle on a refactor.  

I am pretty happy with how things are looking so I would not prefer C.

cis/processor.py
andrewkrug: I see it now... D'oh.

cis/libs/validation.py
gdestuynder: @andrewkrug i think the print is still in there )probably other places too)
using traceback gives you the full traceback instead of just the exception name which is generally very helpful when debugging.

I would recommend dropping all print() outside of tests and either adding or enhance logger code to log the exception traceback wherever there is a catchall (ie `except Exception as e: ...`)

cis/plugins/validation/mozilliansorg_publisher_plugin.py
gdestuynder: @andrewkrug  I will be unavailable for most of next week - my vote would be to let mozilliansorg create users when they dont exist only, and until auth0/workday publishers exist. Happy to talk tomorrow though.

cis/libs/streams.py
johngian: Pleaes file a bug then.

johngian:
There are a lot of replies to comments that suggest fixing stuff in the future. If we want to unblock this PR without them fixed please open up issues.

For example issues that need attention at some point (either now or in the future):

* Logging
* Wrong handling of exceptions
* Subclassing instead of string checking for states
* Env vars for APEX

If there are more feel free to add.

PR #72, opened 2017-08-03
[DRAFT] Add descriptions and some formats to the Profile schema

djmitche:
I've written up some descriptions for schema fields -- and added a few formats where they were clear.

This is all just guesswork -- and where I've written "UNKNOWN" I was unable to guess.  Please comment to let me know where I've gotten it wrong, and I will update.

### DIFF
--- a/cis/schema.json
+++ b/cis/schema.json
@@ -1,5 +1,6 @@
 {
-    "title": "CIS schema",
+    "title": "CIS User Profile",
+    "description": "A user profile for a user managed by CIS.  Objects of this type are generated from information provided by identity providers (IdP's).\n\nFields identified as authoritative can be relied on for identification and access control purposes; fields identified as informational should not be relied on for crticical operations but may be used to support interaction with the user.",
     "type": "object",
     "additionalProperties": false,
     "required": [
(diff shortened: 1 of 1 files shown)


djmitche:
It looks like this has been moved to `cis/plugins/validation/schema.json`.  I don't know why git couldn't follow the move.  I'll rebase once I've gotten some comments.  @gdestuynder - any comments?

djmitche:
@gdestuynder

djmitche:
damn :cat:

gdestuynder:
Thanks for this PR, this is quite useful

I've edited it to reflect how fields are used in some cases (or fixed the `UNKNOWN`s)
I've also removed the wording related to 'authoritative' vs 'informational' as this would not currently reflect how the profile is generated.
All values come for various user management interfaces (workday, mozillians.org, ldap mainly). We call these CIS providers and they may only touch certain attributes, based on validation rules that we maintain in code (they are not expected to change often).

Finally, the file has moved to `https://github.com/mozilla-iam/cis/blob/master/cis/plugins/validation/schema.json` hence the conflict

gdestuynder:
manually merging with conflict resolution

cis/schema.json
```diff
@@ -233,15 +243,15 @@
                     "lastUsed": {
                         "type": "string",
                         "format": "date-time",
-                        "description": "UNKNOWN"
+                        "description": "Date at which this group membership last granted access to a resource (generally, an RP) for the user. This may be used to automatically expire and deny access when the lastUsed attribute is older than 3 month, for example."
```
djmitche: Is there a defined way to update this date?  Or is that something that's only handled for very specific cases (I'm assuming VCS access..).

cis/schema.json
```diff
@@ -233,15 +243,15 @@
...
                         },
                     "name": {
                         "type": "string",
-                        "description": "Name of the group, as given by the IdP."
+                        "description": "Name of the group."
```
djmitche: Does this follow the same format (`provider_name`)?

cis/schema.json
```diff
@@ -233,15 +243,15 @@
...
                         },
                     "uuid": {
                         "type": "string",
-                        "description": "Reserved; profile providers should not use this value."
+                        "description": "Unique identifier for the authoritative group. Profile consumers should not try to interpret this value in any way."
```
djmitche: If I were building a document editing tool and wanted to be able to associate groups with access to a document, would this be the right value to store in my access-control table, rather than name?

cis/schema.json
```diff
@@ -29,54 +29,54 @@
...
-            "description": "(Informational) The primary email address at which the user may be reached.  Profile consumers may use this email to communicate with the user, but should not use it as a stable identifier for the user."
+            "description": "The primary email address at which the user may be reached.  Profile consumers may use this email to communicate with the user, but should not use it as a stable identifier for the user."
         },
         "emails": {
             "type": "array",
-            "description": "(Informational) All email addresses associated with this user's profile. Order is undefined.",
+            "description": "All email addresses associated with this user's profile. Order is undefined.",
             "items": {
                 "type": "object",
                 "properties": {
```
djmitche: It might be worth adding `required` and `additionalProperties` here (and for the other arrays of sub-objects), too.

cis/schema.json
```diff
@@ -29,54 +29,54 @@
     "properties": {
         "user_id": {
             "type": "string",
-            "description": "(Authoritative) The user ID as defined by the underlying IdP. Profile consumers should not try to interpret this value in any way, but may use it as a stable identifier for this user as identified by this IDP."
+            "description": "The unique user identifier. Profile consumers should not try to interpret this value in any way."
```
gdestuynder: i don't want readers to infer that this is to be used as the user identifier for users to login, i.e.
Login: blah|LDAP|xxx
Password: xxxx

Because the user_id is not very human friendly compared to emails. It's ok to use it internally where needed though.

cis/schema.json
```diff
@@ -29,54 +29,54 @@
...
         },
         "timezone": {
             "type": "string",
-            "description": "(Informational) Timezone name for this user's typical location, e.g., 'Europe/London'."
+            "description": "Preferred timezone name for this user's location, e.g., 'Europe/London'."
         },
         "active": {
             "type": "boolean",
-            "description": "(Authoritative) If false, all access for this user should be denied. Profile consumers MUST check this value and disregard the remainder of the profile if false."
+            "description": "If false, all access for this user should be denied. Profile consumers MUST check this value and disregard the remainder of the profile if false."
```
gdestuynder: when querying the CIS API in the future - the point is that it is stored that way in the identity vault (ie the database of user profiles) thus anything that may read the profile would get "false" returned. Right now, this is only "internal" CIS code. Auth0 will return an HTTP error code.

cis/schema.json
```diff
@@ -29,54 +29,54 @@
...
         "created": {
             "type": "string",
             "format": "date-time",
-            "description": "(Authoritative) Date of profile creation"
+            "description": "Date of profile creation"
         },
         "userName": {
             "type": "string",
-            "description": "UNKNOWN"
+            "description": "Human-readable login name or identifier for this user. For a unique identifier, use `user_id` instead."
```
gdestuynder: See above comment about `user_id`

cis/schema.json
```diff
@@ -185,6 +194,7 @@
         },
         "PGPFingerprints": {
             "type": "array",
+            "description": "All PGP fingerprints associated with this user's profile such as `bob` or `mary`. Order is undefined.",
```
gdestuynder: :) if you pr, please fix

cis/schema.json
```diff
@@ -206,22 +216,22 @@
...
         },
         "shirtSize": {
             "type": "string",
-            "description": "(Informational) Shirt size for this user. Format undefined."
+            "description": "Shirt size for this user. Format undefined."
         },
         "groups": {
             "type": "array",
-            "description": "(Informational) Deprecated. Use `authoritativeGroups` instead.",
+            "description": "Informational group membership for the user. Non-prefixed groups are what is known as Mozilla LDAP groups. Groups prefixed by a provider name and underscore are provided by a specific group engine.\nFor example `providername_groupone` is provided by `providername`.\nThese groups are provided `as-is` and you should ensure it contains the members you expect.",
```
gdestuynder: `groups` are used for access control. how much of a guarantee groups are properly *managed* is up to you, authoritative groups have a better guarantee. Both are needed in practice.

cis/schema.json
gdestuynder: So far we've only planned to update this date through CIS, by Auth0, on-access. VCS still uses an LDAP write to a specific LDAP attribute at this time. There could be other publishers such as a VCS in the future though, effectively replacing the LDAP attribute write.

cis/schema.json
gdestuynder: No - while this structure (`authoritativeGroups`) is present it's not yet used - but we currently do not plan to namespace publishers within it. We had a previous iteration where all groups were namespaced (via a json structure, not a string 'hack'/underscore) but we realized it was currently too unrealistic. This is also why, at this time, only Auth0 is planned to write here.

cis/schema.json
gdestuynder: if you wanted to use these groups, then yes. In most cases you'll want to use `groups` though.

There were many questions about `groups` vs `authoritativeGroups` and ideally I/we should make a little doc about it (descriptions are a little short to tell the whole story).

Another attempt through comments though:

- Auth0 (the `access provider`) is expected to maintain and enforce `AuthoritativeGroups`. This means it may prevent a user to login to an RP, regardless of the RP's opinion on this. While **all** RPs have an associated `AuthoritativeGroups` attribute, not all of them prevent access (basically any RP set to "allow anyone in" does not prevent access, it just updates attributes)
- RPs *may* use `AuthoritativeGroups` values if they wish, though Auth0 already did the access validation on their behalf
- RPs *should* use `groups` (any, or combination of groups, as the RP sees fit) with full knowledge that the groups are only as reliable as the people managing them. eg:

  - if there is a group called "mozilliansorg_pplwholikechocolate" managed by a person in Japan that maintain a directory of friends who like chococolate and you decide to use it, it's only as reliable as what this person decides the members are.
  - If you create your **own** "mozilliansorg_pplthatitrust" (mozillians.org) or "pplthatitrust" (LDAP manage) then it's up to you.
  - If you use "workday_staff", these are managed by Mozilla HR, though they currently do not have the same guarantees as `AuthoritativeGroups`, but they hopefully and eventually will. If we get there, there will most likely be a "publisher" attribute in the `AuthoritaiveGroups` (adding being much easier than removing/changing attributes)

cis/schema.json
djmitche: Interesting.  If the set of `authoritativeGroups` is small enough, perhaps each known group could be enumerated in the schema, with suitable descriptions?

PR #74, opened 2017-08-17
address first time new user condition where user_id may be none

### LINKED ISSUE #73: If user does not exist in identity vault catch exception and return
The following dynamodb call is failing some times:

An error occurred (ValidationException) when calling the GetItem operation: One or more parameter values were invalid: An AttributeValue may not contain an empty string: ClientError
Traceback (most recent call last):
File "/var/task/main.py", line 42, in handle
result = p.run()
File "/var/task/cis/processor.py", line 63, in run
return(self._publish_to_stream(self._validator_stage()))
File "/var/task/cis/processor.py", line 69, in _validator_stage
profile_data=self.decrytped_profile
File "/var/task/cis/user.py", line 39, in retrieve_from_vault
response = self.dynamodb_table.get_item(Key=user_key)
File "/var/task/boto3/resources/factory.py", line 520, in do_action
response = action(self, *args, **kwargs)
File "/var/task/boto3/resources/action.py", line 83, in __call__
response = getattr(parent.meta.client, operation_name)(**params)
File "/var/task/botocore/client.py", line 312, in _api_call
return self._make_api_call(operation_name, kwargs)
File "/var/task/botocore/client.py", line 601, in _make_api_call
raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the GetItem operation: One or more parameter values were invalid: An AttributeValue may not contain an empty string

andrewkrug:
Closes #73

### DIFF
--- a/cis/user.py
+++ b/cis/user.py
@@ -36,7 +36,12 @@ class Profile(object):
             self._connect_dynamo_db()
 
         user_key = {'user_id': self.profile_data.get('user_id')}
-        response = self.dynamodb_table.get_item(Key=user_key)
+
+        if user_key is not None:
+            response = self.dynamodb_table.get_item(Key=user_key)
+        else:
+            response = None
+
         self.profile_data = response
 
         return response


gdestuynder:
Note this allows mozillians.org to create users (which is what we want until we have more providers, then we don't)

PR #77, opened 2017-09-07
Only allow this publisher to set `groups` data for user

gdestuynder:
accounts/profiles that were initiated by an IdP that we enforce MFA for,
namely GitHub (github) and LDAP (ad)

### DIFF
--- a/cis/plugins/validation/mozilliansorg_publisher_plugin.py
+++ b/cis/plugins/validation/mozilliansorg_publisher_plugin.py
@@ -36,7 +36,14 @@ def run(publisher, user, profile_json):
         'groups'
     ]
 
-    # Check the easiest case.  None type.
+    # Validate that only whitelisted accounts/profiles issued from vetted IdPs (generally, the ones enforcing MFA)
+    # can get groups assigned as these are used for access control
+    whitelist_idp_with_enforced_mfa = [
+            'github|',  # GitHub has a rule in Auth0 to enforce MFA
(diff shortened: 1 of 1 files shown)


cis/plugins/validation/mozilliansorg_publisher_plugin.py
```diff
@@ -57,6 +64,14 @@ def run(publisher, user, profile_json):
     old_groups = user.get('groups')
     new_groups = profile_json.get('groups')
 
+    for profile_idp in whitelist_idp_with_enforced_mfa:
+        if profile_json.get('user_id').startswith(profile_idp):
```
akatsoulas: Shouldn't this be `if not`?

cis/plugins/validation/mozilliansorg_publisher_plugin.py
```diff
@@ -57,6 +64,14 @@ def run(publisher, user, profile_json):
...
+            if new_groups and len(new_groups) != 0:
```
akatsoulas: probably you don't need the `len(new_groups)`. if the first check, returns true, that means that the list is not empty and we have groups.

cis/plugins/validation/mozilliansorg_publisher_plugin.py
gdestuynder: yup ;-)

akatsoulas:
I thought I had already approved this PR. Sorry for blocking this :|

PR #79, opened 2017-09-20
Add schema field to support tags.

johngian:


cis/plugins/validation/schema.json
```diff
@@ -271,6 +271,13 @@
             "type": "string",
             "description": "Shirt size for this user. Format undefined."
         },
+        "tags": {
+            "type": "array",
+            "description": "General information about users tags. Tags prefixed by a provider name are provided by a specific tag origin.\nFor example `providername_l10n` is provided by `providername`.",
```
akatsoulas: Why not use something similar to groups? It might be confusing to have `mozillians_l10n` for tags and `nda` (without the prefix) for groups.

johngian: both tags and groups are prefixed

johngian:
It's not blocking us at the moment so lets wait for @gdestuynder to take a look as well.

gdestuynder:
@akatsoulas I think we also need the validation rule so that your publisher is allowed to write the tag.

cis/plugins/validation/schema.json
gdestuynder: we call them 'publisher' everywhere else so i'd call it `publishername_` here and elsewhere in the schema (oversight)
we can fix that as a diff pr though, ill open an issue

johngian:
@gdestuynder PR updated with some minor refactoring. How can i test this? My concern is that our current objects are not having a ``tags`` attribute and this is going to fail on ``tags.get()``.

gdestuynder:
tests would be good regardless (such as ensuring the function always works for refusing "wrong groups" for ex)
the code itself otherwise runs at publish time, which means once pushed all profiles need to have 'tags'
either we sync the release between mozillians.org and CIS, or we introduce version numbers to the profile so that CIS knows what to validate per version cc @andrewkrug for input

johngian:
We can always run a migration command to update all our current entries to have empty ``tags: []``

gdestuynder:
would be alright
@johngian  @akatsoulas another thought:

we use publisher prefixes in `user.groups` for compatibility, but we don't really need to do that elsewhere, including `tags` imo. Basically, that'd make Mozillians.org the only publisher allowed to write to `tags` and that's it (we still have use the `mozilliansorg_..` style check to be a function that we can use in each group publisher regardless so no work is lost)

mainly, if we were to cleanly namespace otherwise we would do it with json, such as:
```
user.tags: [
 {'publisher': 'mozilliansorg', 'value': 'blah'},
...
]
```

johngian:
Reverted it back to a similar state. I added a comment that this refers only to Mozillians.org data.

gdestuynder:
r+

PR #86, opened 2017-10-18
Workaround for DynamoDB `None` values limitation

johngian:
* Given a dictionary `d` recursively update all None values with empty string

### DIFF
--- a/cis/publisher.py
+++ b/cis/publisher.py
@@ -77,6 +77,22 @@ class ChangeDelegate(object):
             'signature': signature
         }
 
+    def _nullify_empty_values(self, data):
+        """
+        Recursively update None values with empty string.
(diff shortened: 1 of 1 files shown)


cis/user.py
```diff
@@ -56,8 +56,26 @@ def store_in_vault(self):
...
+
+        def update_values(d):
+            """Recursively update None values with empty string."""
+            new = {}
+            for k in d.keys():
+                v = d[k]
+                if isinstance(v, dict):
+                    v = update_values(v)
+                if v is None:
+                    new[k] = ''
```
andrewkrug: Please change to `Null` we can revisit this choice late 2018 when DynamoDb supports empty string type.

PR #93, opened 2018-01-12
Publishers always send full profile for existing users.

andrewkrug:
The following additional method will do the following:

## The what
1. Publisher constructs profile change.
2. During preparation publisher object will take the RPs client ID / client secret and request a bearer token from auth0.  This token will be used to attempt retrieval of a JSONified profile from the personAPI.  
3. The groups will be added or discarded based on publisher_id.  This means that all publishers must adhere to the convention of matching publisher ids to group prefix. 
4. A combined group list is put into the profile ( pre-encryption ) 

## The how

In order to deploy this PR the following will need to take place.

1. Mozillians.org changes publisher_id to mozilliansorg 
2. Mozillians.org adds the required parameters to the django settings related to CIS. 
3. InfoSec authorizes mozillians.org RP ClientID with scope access to call person API. 
4. Testing in staging with auth0 dev


cis/publisher.py
```diff
@@ -118,6 +124,42 @@ def _validate_profile_data(self):
...
+            }
+        )
+
+        # Retrieve the profile from the CIS API
+        vault_profile = person.get_userinfo(self.profile_data.get('user_id'))
+
+        if vault_profile is not None:
+
+            logger.debug('Vault profile retreived for existing vault user: {}'.format(vault_profile.get('user_id')))
+            publisher_groups = self.profile.get('groups')
```
johngian: Where is ``self.profile`` defined?

johngian: Did you mean ``self.profile_data`` ?

cis/publisher.py
```diff
@@ -118,6 +124,42 @@ def _validate_profile_data(self):
...
+                    reintegrated_groups.append(group)
+
+            for group in vault_groups:
+                # Trust the data in the vault for all other group prefixes.
+                if group.split('_')[0] != self.publisher.get('id'):
+                    reintegrated_groups.append(group)
+
+            logger.debug('Groups successfully reintegrated. Replacing group list in proposed profile.')
+            # Replace the data in the profile with our reintegrated group list.
+            self.profile['groups'] = reintegrated_groups
```
johngian: Same here.

cis/libs/api.py
```diff
@@ -0,0 +1,53 @@
+import http.client
+import json
+import urllib
+
+
+class Person(object):
+    """Retrieve data from person api as needed.  Will eventually replace Mozillians API"""
```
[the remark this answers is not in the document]
johngian: I guess this is a copy/paste from the profile API right ?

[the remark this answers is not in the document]
andrewkrug: I actually copied and pasted from the SSO Dashboard api class.  I'll remove this.

cis/libs/api.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, person_api_config):
+        """
+        :param person_api_config, a dictionary of configuration information about how to connect to person_api
+        """
+
+        # Audience is either https://person-api.sso.mozilla.com or https://person-api.sso.allizom.org
+        self.audience = person_api_config.get('audience')
```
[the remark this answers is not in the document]
andrewkrug: The audience is part of the OIDC API on the other side.  They don't necessarily have to be the same as the URL but in this case they are.  Though it may not always be that way for all endpoints.

cis/libs/api.py
```diff
@@ -0,0 +1,53 @@
...
+        payload = json.dumps(
+            {
+                'client_id': self.client_id,
+                'client_secret': self.client_secret,
+                'audience': self.audience,
+                'grant_type': 'client_credentials'
+            }
+        )
+
+        headers = {'content-type': "application/json"}
```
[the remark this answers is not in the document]
andrewkrug: Didn't want to add another library.  This code is replicated in the SSO Dashboard.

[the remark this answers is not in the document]
johngian: Sounds good. I use ``requests`` in most of my projects because it considered to be a standard. That looks ok as well.

cis/libs/api.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, person_api_config):
+        """
+        :param person_api_config, a dictionary of configuration information about how to connect to person_api
+        """
+
+        # Audience is either https://person-api.sso.mozilla.com or https://person-api.sso.allizom.org
+        self.audience = person_api_config.get('audience')
+        self.client_id = person_api_config.get('client_id')
+        self.client_secret = person_api_config.get('client_secret')
+        self.oidc_domain = person_api_config.get('oidc_domain')
```
gdestuynder: technically its oauth2 but not/without oidc in this case (nitpick)

cis/libs/api.py
```diff
@@ -0,0 +1,53 @@
...
+                'client_secret': self.client_secret,
+                'audience': self.audience,
+                'grant_type': 'client_credentials'
+            }
+        )
+
+        headers = {'content-type': "application/json"}
+
+        conn.request("POST", "/oauth/token", payload, headers)
+        res = conn.getresponse()
```
gdestuynder: might want some error checking at this level instead of api failures, rate limits, etc (at the minimum http code check)

cis/libs/api.py
```diff
@@ -0,0 +1,53 @@
...
+        return json.loads(data.decode('utf-8'))
+
+    def get_userinfo(self, auth_zero_id):
+        user_id = urllib.quote(auth_zero_id)
+        conn = http.client.HTTPSConnection("{}".format(self.person_api_url))
+        token = "Bearer {}".format(self.get_bearer().get('access_token'))
+
+        headers = {'authorization': token}
+
+        conn.request("GET", "/prod/profile/{}".format(user_id), headers=headers)
```
gdestuynder: i believe thats no longer compatible with our url versioning
we could have a new value called `person_api_version = "v1"` though and autoconstruct the url from that

cis/publisher.py
```diff
@@ -118,6 +124,42 @@ def _validate_profile_data(self):
         # Validate data prior to sending to CIS
         return validation.Operation(self.publisher, self.profile_data).is_valid()
 
+    def _reintegrate_profile_with_api(self):
+        person = api.Person(
+            person_api_config={
+                'audience': self.config('person_api_audience', namespace='cis'),
+                'client_id': self.config('oidc_client_id', namespace='cis'),
+                'client_secret': self.config('oidc_client_secret', namespace='cis'),
+                'oidc_domain': self.config('oidc_domain', namespace='cis'),
```
gdestuynder: nit: technically its just oauth2 not oidc

cis/publisher.py
```diff
@@ -118,6 +124,42 @@ def _validate_profile_data(self):
...
+
+        if vault_profile is not None:
+
+            logger.debug('Vault profile retreived for existing vault user: {}'.format(vault_profile.get('user_id')))
+            publisher_groups = self.profile_data.get('groups')
+            vault_groups = vault_profile.get('groups')
+
+            reintegrated_groups = []
+
+            for group in publisher_groups:
```
gdestuynder: this only reintegrates groups - i understand that this is so that it can be generic, but for example mozillians.org has access to more fields than `user.groups`
say, what about tshirt size?

I'm blocking review on this one (though maybe I missed/misunderstood something) as it might break stuff

cis/libs/api.py
andrewkrug: fixed

cis/libs/api.py
andrewkrug: fixed

cis/libs/api.py
andrewkrug: done

cis/publisher.py
andrewkrug: fixed

cis/publisher.py
andrewkrug: mozillians.org is authoritative for that data so... no need to reintegrate.

gdestuynder: hah my bad i was confused

cis/publisher.py
```diff
@@ -118,6 +124,43 @@ def _validate_profile_data(self):
...
+                'person_api_version': self.config('person_api_version', namespace='cis')
+            }
+        )
+
+        # Retrieve the profile from the CIS API
+        vault_profile = person.get_userinfo(self.profile_data.get('user_id'))
+
+        if vault_profile is not None:
+
+            logger.debug('Vault profile retreived for existing vault user: {}'.format(vault_profile.get('user_id')))
```
gdestuynder: typo on retrieved ;-)

PR #97, opened 2018-01-25
Continuing work on full profiles

andrewkrug:
This PR adds basic sign/verify support to CIS as well as allows mozillians to push users to the identity vault if no user exists.

### DIFF
--- a/cis/plugins/validation/mozilliansorg_publisher_plugin.py
+++ b/cis/plugins/validation/mozilliansorg_publisher_plugin.py
@@ -72,8 +72,8 @@ def run(publisher, user, profile_json):
     # group sub structure looks like: user.groups = [ 'group_from_ldap1', 'moziliansorg_mytestgroup', ...]
     prefix = 'mozilliansorg_'
 
-    old_groups = user.get('groups')
-    new_groups = profile_json.get('groups')
+    old_groups = user.get('groups', [])
+    new_groups = profile_json.get('groups', [])
 
     for profile_idp in whitelist_idp_with_enforced_mfa:
         if not profile_json.get('user_id').startswith(profile_idp):
--- a/cis/plugins/validation/mozilliansorg_publisher_plugin.py
+++ b/cis/plugins/validation/mozilliansorg_publisher_plugin.py
@@ -6,11 +6,11 @@ utils.StructuredLogger(name=__name__, level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 # Allows user creation by this publisher.
-CAN_CREATE_USER = True
+CAN_CREATE_USER = True  # XXX TBD turn this back to false when there is another method of user provision.
 ENFORCE_ATTR_WHITELIST = False
 
 
-def run(publisher, user, profile_json):
+def run(publisher, vault_json, profile_json):
     """
     Check if a user mozillian's profile properly namespaces mozillians groups
     and whitelists which fields mozillians.org has authority over.
@@ -52,7 +52,7 @@ def run(publisher, user, profile_json):
     ]
 
     # Check the easiest case. None type.
-    if user is None and CAN_CREATE_USER is False:
+    if vault_json is None and CAN_CREATE_USER is False:
         logger.exception('permission denied: publisher {} attempted to modify user that does not exist'
                          ' in the identity vault'.format(publisher))
         return False
@@ -60,9 +60,9 @@ def run(publisher, user, profile_json):
     # Attr whitelist is currently disabled as mozilliansorg needs to be able to create new Profiles
     # XXX TBD This should pull back a profile from person api for this comparison
     if ENFORCE_ATTR_WHITELIST is True:
-        for attr in user:
+        for attr in vault_json:
             if attr not in whitelist:
-                if profile_json.get(attr) != user.get(attr):
+                if profile_json.get(attr) != vault_json.get(attr):
                     logger.exception('permission denied: publisher {} attempted to modify user attributes it has no'
                                      'authority over'.format(publisher))
                     return False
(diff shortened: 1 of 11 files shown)


cis/plugins/validation/mozilliansorg_publisher_plugin.py
```diff
@@ -5,6 +5,10 @@
 utils.StructuredLogger(name=__name__, level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+# Allows user creation by this publisher.
+CAN_CREATE_USER = True
+ENFORCE_ATTR_WHITELIST = False
+
 
 def run(publisher, user, profile_json):
```
gdestuynder: we probably should call `user` something like `user_from_vault` (i got confused even thus i wrote the contents of the plugin initially)

cis/plugins/validation/mozilliansorg_publisher_plugin.py
```diff
@@ -5,6 +5,10 @@
 utils.StructuredLogger(name=__name__, level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+# Allows user creation by this publisher.
+CAN_CREATE_USER = True
+ENFORCE_ATTR_WHITELIST = False
```
gdestuynder: add XXX here too saying we need to turn this back on later (in fact mainly here)

cis/processor.py
```diff
@@ -32,6 +34,11 @@ def run(self):
         # Determine what stage of processing we are in and call the corresponding functions.
         self.decrytped_profile = json.loads(self._decrypt_profile_packet())
```
gdestuynder: nit that typo ;-)

cis/processor.py
```diff
@@ -32,6 +34,11 @@ def run(self):
...
 
+        if self.signature == {} or self.signature is None:
+            self.signature = self.decrytped_profile.get('signature')
+
+        self._verify_signature()
```
gdestuynder: should we fail here, or at least warn for now if this returns False?

cis/processor.py
```diff
@@ -62,6 +88,11 @@ def __init__(self, boto_session, publisher, signature, encrypted_profile_data):
     def run(self):
         logger.info('Attempting to load stage processor logic: {}'.format(self.stage))
         self.decrytped_profile = json.loads(self._decrypt_profile_packet())
+
+        if self.signature == {} or self.signature is None:
+            logger.info('Signature not set on object. Retreiving from profile packet.')
```
gdestuynder: nit typo

cis/processor.py
```diff
@@ -71,16 +102,32 @@ def _validator_stage(self, kinesis_client=None):
...
+            result = validation.Operation(
+                publisher=self.publisher,
+                profile_data=self.decrytped_profile,
+                user=self.user
+            ).is_valid()
+        else:
+            result = False
+
+        logger.info(
+            'Result of the validation operation for user_id: {} is {}'.format(
```
gdestuynder: 'Signature verification failed' please - make it obvious

cis/publisher.py
```diff
@@ -162,13 +183,44 @@ def _reintegrate_profile_with_api(self):
...
+                    self.publisher.get('id')
+                )
+            )
+
+            self.stats(vault_profile, self.profile_data)
             # Replace the data in the profile with our reintegrated group list.
             self.profile_data['groups'] = reintegrated_groups
 
-    def _generate_signature(self):
+    def stats(self, current_user_profile, proposed_change):
```
gdestuynder: what is this used for?

cis/publisher.py
```diff
@@ -162,13 +183,44 @@ def _reintegrate_profile_with_api(self):
...
+            logger.info(
+                'The net group change for user_id: {} is {} number of groups.'.format(user_id, net_change)
+            )
+
+    def _generate_signature(self, profile_data):
         # If signature doesn't exist attempt to add one to the profile data.
-        return {}
+        o = crypto.Operation()
+        sig = base64.b64encode(o.sign(plaintext=profile_data))
+        logger.info('Signature generated for user_id: {}. The sigature is {}'.format(
```
gdestuynder: typo

cis/publisher.py
```diff
@@ -55,24 +56,28 @@ def _connect_aws(self):
...
+            )
             return result
 
     def _get_event_dict(self):
-
-        encrypted_profile = str(base64.b64encode(self._prepare_profile_data()))
-        signature = self._generate_signature()
+        profile_data = self._prepare_profile_data()
+        encrypted_profile = str(base64.b64encode(profile_data))
+        signature = self._generate_signature(json.dumps(self.profile_data))
```
gdestuynder: should check if `_generate_signature` failed here, or in `_generate_signature`

cis/publisher.py
andrewkrug: It generates some log messages about how many groups were added or removed.  I'm going to turn this into a cloudwatch metric for mass group add / drops for anomalies.

gdestuynder: oh cool

cis/processor.py
andrewkrug: We actually will fail it.  I put signature verification in this parent object so it would be accessible in all of the objects that take OperationDelgate as a class.  Basicall if you look down in line 105 this is where that "fail and log behavior" is in code.

cis/processor.py
andrewkrug: Actually ... this logger is one indent out of "sig verification failed."  I'm adding another log statement to this block though that'll live in that else: on line 118

cis/publisher.py
andrewkrug: Signature generation is non-optional the way pykmssig works.  If a signature can not be generated it will raise.

cis/plugins/validation/mozilliansorg_publisher_plugin.py
```diff
@@ -5,8 +5,12 @@
 utils.StructuredLogger(name=__name__, level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+# Allows user creation by this publisher.
+CAN_CREATE_USER = True  # XXX TBD turn this back to false when there is another method of user provision.
+ENFORCE_ATTR_WHITELIST = False
```
gdestuynder: and that to True!

PR #98, opened 2018-03-01
added feature to pass assumed role down into pykmssig when present

andrewkrug:


cis/processor.py
```diff
@@ -59,7 +59,12 @@ def _verify_signature(self):
             return False
 
     def _decrypt_and_verify(self):
-        o = crypto.Operation()
+        if self.boto_session is None:
+            self._connect_aws()
+
```
gdestuynder: i guess another possible idom is `self.boto_session = self._connect_aws()` or similar (reduces reading complexity)

andrewkrug: We use an object attribute like this all over because boto_session gets reused.  self._connect_aws() could be made to also return the session though.

PR #106, opened 2018-03-13
Deferred bearer token

andrewkrug:


cis/libs/api.py
```diff
@@ -42,15 +44,38 @@ def get_bearer(self):
         )
 
         headers = {'content-type': "application/json"}
-        conn.request("POST", "/oauth/token", payload, headers)
 
-        res = conn.getresponse()
-        if res.status == 200:
-            data = res.read()
-            return json.loads(data.decode('utf-8'))
+        if self._refresh():
```
gdestuynder: this could probably use https://github.com/mozilla-iam/authzerolib/blob/master/authzero.py

andrewkrug: Adding dependencies to this is a little rough but let's talk about refactoring for that.

PR #108, opened 2018-03-26
add proposal for new user profile schema + schema validator

gdestuynder:
Note that the validator now uses jsonschema.org's draft 7 (latest) though it can be down-converted (python's validator supports up to draft 4 only...)

Note also that this does not include https://github.com/lbovet/docson generation, though if we adopt this we'll want to add it - most likely as GitHub pages for this repos.

Finally, I'm pinging a few more folks here that participated to our tech alignment for feedback. @johngian @akatsoulas @jonasfj @fiji-flo @jdow

Here are rendered versions of the files for easier consumption:

- https://github.com/gdestuynder/cis/blob/profile/docs/rfcs/UserProfilesv2.md
- https://github.com/gdestuynder/cis/blob/profile/docs/rfcs/UserProfilesv2_schema.json (you may load that in https://www.jsonschemavalidator.net/ or similar - otherwise the `description` fields are most likely what you care about most)

### DIFF
--- a/docs/rfcs/UserProfilesv2.md
+++ b/docs/rfcs/UserProfilesv2.md
@@ -49,8 +49,6 @@ IAM Goals:
   "created": "2017-03-09T21:28:51.851Z",
   "userName": "lmcardle@mozilla.com",
   "displayName": "Leo McArdle",
-  "firstName": "Leo",
-  "lastName": "McArdle",
   "preferredLanguage": "en_US",
   "primaryEmail": "lmcardle@mozilla.com",
   "identities": [
--- a/docs/rfcs/UserProfilesv2.md
+++ b/docs/rfcs/UserProfilesv2.md
@@ -1,0 +1,145 @@
+# User Profiles (v2)
+
+This document describes the Mozilla IAM user profiles version 2.
+
+IAM Goals:
+- MUST stay compatible with v1 (current profiles)
+- Add structured access information that can be reliably queried
+- Integrate the [Automatic Access Expiration](AutomaticAccessExpiration.md) to the access information
+- Allow for identity information to be transported from social authentication providers to our profile
+- Standardize documentation around the profile
+- Allow for a standardized API on top of the user profile data ([Person-API](https://github.com/mozilla-iam/person-api))
+- Version the profile for possible future updates
+
+## Use cases
+
+1. Machines such as TaskCluster want to reliably query Person-API and get authoritative answers about the user's group
+   membership. The groups must be easy to query with no ambiguity (i.e. no string-namespacing).
+2. User managing GitHub wants to access GitHub user_ids (not email) from logged in users in order to query the GitHub
+   API and get relevant data about the user.
+3. The access provider wants to expire access when it is no longer used (Automatic Access Expiration).
+4. Security team wants to assert that access is lost to certain services (RPs) when they're no longer used by the person
+   who used to have access.
+
+## Profile proposal
+
+**Main changes**
+
+1. `_schema` field contains the URL to the json-schema.org validation file. The validation file itself contains the
+   revision-equivalent, as well as the environment the schema is written for directly in the URL. This field also
+   mirrors it, eg: "person-api.sso.allizom.org" => dev. environment, "v2" is version 2.
+2. `emails` structure becomes `identities`. This is a dangerous change as it removes a previously existing field. After
+   careful examination we have determined that this field is very unlikely to be currently used and have decided to
+   replace it. Identities can contain more than emails, and in particular, can contain an upstream provider's `user_id`.
+   Note that a refactor of Person-API and certain CIS libraries is required for this change.
+3. `groups` are unchanged for compatibility.
+4. `accessInformation` is a new structure that combines `authoritativeGroups` (this structure is known with a very high
+   degree of certainty to be unused) and `groups` data. Most sub-fields of that structure are optional, and it's
+   organized per "access information publisher".
+
+### Example profile:
+
+```
+{
+  "_schema": "https://person-api.sso.mozilla.com/profile/v2/schema",
+  "user_id": "ad|Mozilla-LDAP-Dev|lmcardle",
+  "timezone": "Europe/London",
+  "active": true,
+  "lastModified": "2017-03-09T21:28:51.851Z",
+  "created": "2017-03-09T21:28:51.851Z",
+  "userName": "lmcardle@mozilla.com",
+  "displayName": "Leo McArdle",
+  "firstName": "Leo",
+  "lastName": "McArdle",
+  "preferredLanguage": "en_US",
+  "primaryEmail": "lmcardle@mozilla.com",
+  "identities": [
+    {
+      "email": "lmcardle@mozilla.com",
+      "verified": true,
+      "primary": true,
+      "lastModified": "2017-03-09T21:28:51.851Z",
+      "verifier": "MozillaLDAP",
+      "user_id": "dn=mozilla.com,cn=lmcardle"
+    },
+    {
+      "email": "leomcardle@gmail.com",
+      "verified": true,
+              "primary": false,
+      "verifier": "github-oauth2",
+      "lastModified": "2017-03-09T21:28:51.851Z",
+      "user_id": "834847"
+    }
+  ],
+  "phoneNumbers": [
+    "+4958339847"
+  ],
+  "uris": [
+    "https://blog.example.net"
+  ],
+  "nicknames": [
+    "leo"
+  ],
+  "SSHFingerprints": [
+    "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCiAoThvwWQaiTLdkGVvUKbkhmNX9X+cvJZRKnoiv7iGHBKTw4flcTSkwyJQzXTep8R"
+  ],
+  "PGPFingerprints": [
+    "-----BEGIN PGP PUBLIC KEY BLOCK-----\n\nmQINBE94eWwBEADjlvvF8HERvp.....=A0dq\n-----END PGP PUBLIC KEY BLOCK-----\n"
+  ],
+  "picture": "https://s.gravatar.com/avatar/ec6e85d15f8411d32f97f5d8a4eab2d3?s=480&r=pg&d=https%3A%2F%2Fcdn.auth0.com%2Favatars%2Flm.png",
+  "shirtSize": "M",
+  "groups": [
+    "0A_legacy_structure_only_please_use_accessInformation_instead",
+    "mozilliansorg_bar",
+    "mozilliansorg_foo",
+    "hris_foo",
+    "ldapfoo"
+  ],
+  "accessInformation": {
+    "ldap": [
+      {
+        "name": "ldapfoo"
+      },
+      {
+        "name": "vpn_default"
+      }
+      ],
+    "mozilliansorg": [
+      {
+        "name": "nda"
+      },
+      {
+        "name": "nda"
+      }
+      ],
+    "hris": [
+      {
+        "name": "costcenter",
+        "value": "1420"
+      },
+      { 
+        "name": "workertype",
+        "value": "employee"
+      },
+      {
+        "name": "egencia",
+        "value": "uk"
+      },
+      {
+        "name": "department",
+        "value": "IT"
+      }
+      ],
+    "accessprovider": [
+      {
+        "created": "2010-01-23T04:56:22Z",
+        "lastUsed": "2010-01-23T04:56:22Z",
+        "name": "mozdef1.private.scl3.gmail",
+        "value": "5a5munnfxYjqkaN0su1Kl7USxbqkILQN"
+      }
+    ]
+  }
+}
+```
+
+Schema validator: [here](UserProfilesv2_schema.json)
--- a/docs/rfcs/UserProfilesv2.md
+++ b/docs/rfcs/UserProfilesv2.md
@@ -39,6 +39,8 @@ IAM Goals:
 
 ### Example profile:
 
+This is all the profile data available to Mozilla IAM, though RPs may be able to see or query only parts of it.
+
 ```
 {
   "_schema": "https://person-api.sso.mozilla.com/schema/v2/profile",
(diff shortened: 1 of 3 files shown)


docs/rfcs/UserProfilesv2_schema.json
```diff
@@ -0,0 +1,486 @@
+{
+  "$id": "https://person-api.sso.mozilla.com/profile/v2/schema",
+  "type": "object",
```
jonasfj: Often you'll want to specify `"additionalProperties": false,` when you have type object, as it defaults to allowing unspecified properties.

docs/rfcs/UserProfilesv2.md
```diff
@@ -0,0 +1,145 @@
...
+   who used to have access.
+
+## Profile proposal
+
+**Main changes**
+
+1. `_schema` field contains the URL to the json-schema.org validation file. The validation file itself contains the
+   revision-equivalent, as well as the environment the schema is written for directly in the URL. This field also
+   mirrors it, eg: "person-api.sso.allizom.org" => dev. environment, "v2" is version 2.
+2. `emails` structure becomes `identities`. This is a dangerous change as it removes a previously existing field. After
```
jonasfj: You could also just duplicate the property, and document the "emails" property as deprecated...

But it does look like over kill in this case...

docs/rfcs/UserProfilesv2_schema.json
```diff
@@ -0,0 +1,486 @@
+{
+  "$id": "https://person-api.sso.mozilla.com/profile/v2/schema",
```
jonasfj: Is this supposed to be the return value from the person API as well as the auth0 profile?

If the person API isn't in use and you are thinking the person API is how others at mozilla get information about someone who is logged in... Then I would implement the person API as a transform of the auth0 profile (or whatever data sources you have) and exploit the fact that nobody uses it at the moment to reshape the data structure..

Notably: I would expose only information that people request, and that you know they can rely on. So that you API surface is more minimalistic, giving you more freedom in the future.

But this might be out of scope :)

docs/rfcs/UserProfilesv2_schema.json
```diff
@@ -0,0 +1,486 @@
...
+      "default": "Europe/London",
+      "examples": [
+        "Europe/London"
+      ]
+    },
+    "active": {
+      "$id": "/properties/active",
+      "type": "boolean",
+      "title": "The Active Schema ",
+      "description": "If true, the user is valid. If false, the user is considered de-activated and all access MUST be denied. Users MAY be re-activated. Profile consumers MUST check this value.",
```
jonasfj: This seems like a trap, wouldn't it better to return 404 than a document saying active: false?

IMO this should be loudly highlighted in API docs and examples, I doubt we check this property :)

docs/rfcs/UserProfilesv2_schema.json
```diff
@@ -0,0 +1,486 @@
...
+      }
+    },
+    "phoneNumbers": {
+      "$id": "/properties/phoneNumbers",
+      "type": "array",
+      "items": {
+        "$id": "/properties/phoneNumbers/items",
+        "type": "string",
+        "title": "The 0th Schema ",
+        "description": "All phone numbers that are associated with the user profile. Order is undefined.",
```
jonasfj: Are these verified?

jonasfj: Just saying, that's a good thing to document :)

docs/rfcs/UserProfilesv2_schema.json
```diff
@@ -0,0 +1,486 @@
...
+          "items": {
+            "$id": "/properties/accessInformation/properties/mozilliansorg/items",
+            "type": "object",
+            "properties": {
+              "name": {
+                "$id": "/properties/accessInformation/properties/mozilliansorg/items/properties/name",
+                "type": "string",
+                "title": "The Name Schema ",
+                "description": "Access information Mozillians.org (`mozilliansorg_` in `groups`). This list all access groups in Mozillians.org that this user is member of.",
+                "default": "",
```
jonasfj: A "pattern" documenting the characters allows here would be nice!

I know we rely on LDAP and mozillians group names consistenting of certain character classes, ASCII printable I believe (without *).

It would be neat to see these character classes documented and validated.

Probably `"pattern": "^[a-zA-Z0-9_-]+$"` is fairly safe to add here... And it gives some super useful invariants for downstream consumers like me :)

docs/rfcs/UserProfilesv2.md
```diff
@@ -0,0 +1,145 @@
...
+    "leo"
+  ],
+  "SSHFingerprints": [
+    "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCiAoThvwWQaiTLdkGVvUKbkhmNX9X+cvJZRKnoiv7iGHBKTw4flcTSkwyJQzXTep8R"
+  ],
+  "PGPFingerprints": [
+    "-----BEGIN PGP PUBLIC KEY BLOCK-----\n\nmQINBE94eWwBEADjlvvF8HERvp.....=A0dq\n-----END PGP PUBLIC KEY BLOCK-----\n"
+  ],
+  "picture": "https://s.gravatar.com/avatar/ec6e85d15f8411d32f97f5d8a4eab2d3?s=480&r=pg&d=https%3A%2F%2Fcdn.auth0.com%2Favatars%2Flm.png",
+  "shirtSize": "M",
```
claudijd: Does fitted/non-fitted matter here?  I seem to recall this being a thing for some folks.

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: i thought of basically releasing a new profile schema when that's needed - in particular, we sign entire profiles and publishers of data have to be on the same profile version regardless, i.e. not sure how much it gives us to allow additional properties. That said if you've a use case or if i'm misunderstood this please correct me

docs/rfcs/UserProfilesv2.md
gdestuynder: yeah id rather have as little deprecated fields as possible for simplicity, if possible

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: this url is indeed incorrect, so thanks for pointing that out :)
@andrewkrug any idea what url we'd want for the id? It cannot be an endpoint path or we run into the above (where `schema` could be seen as an argument, but its not)

@jonasfj the other points aren't wrong either though it is out of scope for the "full profile". usually i'd think we'd trim the returned valued depending on scopes

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: I think that's a good idea. we'd still store this internally, but not require consumers to check, just return 404. It's also more in line with common practices. cc @andrewkrug

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: Currently we have no verified phone numbers though its technically possible, there's no plan to implement verification. We could have a generic optional attribute though.

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: @nemo @tasos do you have any such filter on group names in mozillians.org right now? I agree it'd be nice to have this and we could standardize on something. LDAP is a-Z0-9 so far, to my knowledge (@jdow?)

docs/rfcs/UserProfilesv2_schema.json
jonasfj: I meant you should disallow additional properties... JSON schema defaults to allowing them..

jonasfj: Not a big thing... Since you're probably not validating against this schema... But if you are... Perhaps you have tests, disallowing additional properties ensure that they don't sneak in :)

docs/rfcs/UserProfilesv2_schema.json
jonasfj: My point is mostly that it would be good to document that the phone number is not verified.

Just in case someone decides to use an API to write text messages or build a call robot :)

It's more of a good practice kind of thing... In reality people probably won't expect it to be validated.

docs/rfcs/UserProfilesv2_schema.json
jonasfj: It doesn't have to be the same for both LDAP and mozillians... That would be great...

But just documenting what character classes are permitted... Probably also documenting the maximum and minimum string lengths...

It's nice to know that my services will never have to deal with the mozillians group named empty string... Or an LDAP group with a name that is 9000 chars long... Those probably won't happen, so let's document the limits as it reduces the corner cases downstream consumers have to deal with.

In reality downstream consumers won't deal with extreme corner cases I outline above. But documenting what they get will it more likely they handle all inputs correctly.

emceeaich:
# Questions and Suggestions

* From the draft, it looks like I have to infer if a person is staff, mozillian under nda, or other by looking at the contents of the `groups` array or the `accessInformation` object
* HRIS block should contain staff member's preferred name and pronouns 
    * Don't import gender but just the person's pronouns (a person could be transitioning gender on the job but still waiting on an official gender marker change, so their gender marker on their driver's license (and in HRIS systems) could be 'Female' but their pronouns are 'they/their') 
    * If none are specified then don't assume "they/them", use "Not specified"
* Mozillians should only keep for preferred name and pronouns 
    * Wallet names (if different than display name) should not be in Mozillians

Asking Tara R. to review from D&I team.

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: technically, we would disallow (like on v1) though I found this:
https://github.com/json-schema-org/json-schema-spec/blob/master/jsonschema-hyperschema.xml#L798-L800
So i haven't done it. I suspect it's not a huge issue either way though yeah, we could still use it to "make sure".

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: going with `https://person-api.sso.mozilla.com/schema/v2/profile` any comments on it?

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: note its a personapi thing so im keeping this in the schema

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: ✓

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: do you know how to do string len? @jonasfj

gdestuynder:
@emceeaich 
Hi!, here some clarifications on the schema in addition to our meeting (also a summary of some things that were said there, just to be sure it's recorded somewhere).
Also, please take these as indicative only and subject to change, be commented on, critized, modified, etc. Hopefully they help to conversation going forward and helps us making the best profile schema we can!

# Some context 
- the profile data is what we store but not necessarily what we expose
- only data that is used or necessary for things to function is stored
- different relying parties (RP, i.e.: web properties) may required different parts of the profile, but not necessarily all of the profile. Due to this, our API classifies data so that we can present different classes of data depending on the RP. Hopefully, we can make good use of this and classify data correctly. @andrewkrug do you think we should have an additional, classification related schema of some kind to make this more clear?

# Action item (there might be more in the future)
- [ ]  With that in mind, we also want to have a "bare minimum profile example" that is sent to RPs when all that is needed is basic user authentication (this is the most common case) as this needs no specific approval to access data in the profile. I've an action item to produce that here (in a subsequent commit).
- [ ] drop `firstName` and `lastName` (see below why)

# Specific answers to the comment above regarding specific field, these being displayed or not to RPs

## Fields/Claims
- `user_id` is generated by Auth0 and we're unfortunately not in control of how this is created. For additional technically reasons we'd actually like to have this control though this is not the case at this time.

- `userName` and `primaryEmail` are in this case decided by LDAP. These could be different, since it's often presented by RPs - though it requires a broader discussion. I don't really have an opinion on it other than "it has to be as stable as possible" but the actual string/email itself does not have to be a real name from a technical point of view. The stability is needed because RPs most often use this as `user_id` even thus it's technically incorrect. Since we don't control RPs (many of them are vendors), stability is needed there so that users get correctly identified.

- `displayName` vs `firstName` and `lastName`:  the 2 fields first and last name are concatenated automatically into display name and all 3 are used to construct the (OpenID Connect Claim)[https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims] and the (SCIM structure)[https://tools.ietf.org/html/rfc7643#section-4.1.1] I believe we might actually be able to drop these without affecting compatibility and only have `displayName` directly construted from a prefered name (which depends on the IdP). Due to this, I will drop them off the schema.

- `nickname` is entierely optional and basically another `displayName`, which defaults to empty. This allows to storing additional display names for things like IRC, Slack, etc. Technically, it could be dropped, though so far we've seen it used in Phonebook for example. We *might* use `identities` instead though, maybe? (note: this is also a SCIM inherited field)

## Capabilities of the profile
### Staff vs non-staff and other HRIS capabilities
This is currently passed as a `hris_...` group in `user.group` today, but would be passed both there (for compatiblity) and in `user.accessInformation{}.hris[is_staff: True]` in profile v2. nda'd is a separate attribute, see below.
This structure would also contain the manager name, though this will not be passed in `user.group` and is not passed there today due to both technical limitations in `user.group`, and the fact that its not yet needed for any RP. We have use cases where it's needed though (I think you have one of these use case we have recorded, though we also need it for org-chart purposes)

### NDA'd Mozillian
This is current passed as `mozilliansorg_nda` group in `user.group` today but would be passed both there (for compatibility) and in `user.accessInformation{}.mozilliansorg['nda']` in profile v2.
It is possible for someone to be both Staff and NDA'd Mozillian, technically speaking.

### Bugzilla id
The Bugzilla id is not currently recorded, but will be available in profile v2 as soon as the Bugzilla integration is done. This will be available in `user.identities`, for example, `user.identities[].{'email': 'test@test.com', ... 'user_id': 'bugzilla-user-id-here', 'verifier': 'bugzilla-mozilla-com', ..}`
The important check here is that the `verifier` field is BMO. We do not currently have the Bugzilla integration so the `verifier` name has not yet been reserved, but it will be. We're working on this with @dylanwh

### Person's legal ID
As long as legal and HRIS are ok with us using the prefered name in `displayName` we could make sure that the LDAP DN uses the same preferred name, which in turn will ensure the `user_id` and `primaryEmail` do not echo the person legal ID/legal name if they have chosen not to display it. One limitation is that changing this requires recreating the user entierely (which happens sometimes for various reasons, including some you have mentionned). This is unfortunately a technical limitation at this time, as per the `user_id` section above. LDAP suffers from the same limitation as Auth0 in fact, though we have more control on the LDAP side and might eventually find a way to do this differently.

### Gender
We do not currently record gender or gender pronouns as we have no use case for it. This means we also don't assume it of course. Ideally, we'd rather not record it as per our principle of recording only the data we have use cases for. That being said, one may include their preferred pronouns in the `displayName` if they wish to advertise them. This seems to work relatively well on some RPs (mainly, this is because even if we recorded it we have no RP that support this at this time) and I've seen some users do it, and this seems to work fine so far.

### Mozillians.org
Mozillians.org is becoming the profile editor for IAM - though it does not mean it needs access to legal names. Currently, we'd have to check what is passed exactly from LDAP but that's a good call out, together with all the `displayName` items in this comment - I think that we should be able to ensure that.

Hope this helps and thanks for the feedback! (there should be some commits following this comments with the changes I can make now)

gdestuynder:
@jdow looking for your re-review of https://github.com/mozilla-iam/cis/pull/108/commits/84ea9d7677aecc26bd2dff48ac0dee70541c8b93 in particular, would to know if you foresee any compatibility issues with this

gdestuynder:
@emceeaich First attempt at minimal profile: https://github.com/gdestuynder/cis/blob/profile/docs/rfcs/UserProfilesv2.md#example-minimum-profile

We still have to test that this actual work, but it's pretty close to what RPs get today (they get `given_name` and `family_name` in addition to that today, though these are taken from `displayName` aka `name` in OIDC)

johngian:
Suggestions from meeting:

* Create an IAM experimental namespace that we allow flexible changes of the fields
* Have a way to allow permission level defined by user (aka mozillians.org permission levels) 
* We are not going to facilitate custom set of permission groups (individual, staff-nda, public)

gdestuynder:
closing in favor of pr #109 which includes all above feedback as well

PR #110, opened 2018-05-10
Profilev2

gdestuynder:
Fixes discussed with gene and andrew
this needs squashing.
https://github.com/gdestuynder/cis/blob/profilev2/docs/rfcs/UserProfilesv2.json
https://github.com/gdestuynder/cis/blob/profilev2/docs/rfcs/UserProfilesv2_schema.json

verify @ https://www.jsonschemavalidator.net/ draft 4-6-7 compatible

### DIFF
--- a/docs/rfcs/UserProfilesv2.json
+++ b/docs/rfcs/UserProfilesv2.json
@@ -332,7 +332,7 @@
       "value": [
         {
           "created": "2010-01-23T04:56:22Z",
-          "lastUsed": "2010-01-23T04:56:22Z",
+          "last_used": "2010-01-23T04:56:22Z",
           "name": "vpn_default",
           "value": null
         }
@@ -353,7 +353,7 @@
       "value": [
         {
           "created": "2010-01-23T04:56:22Z",
-          "lastUsed": "2010-01-23T04:56:22Z",
+          "last_used": "2010-01-23T04:56:22Z",
           "name": "nda",
           "value": null
         }
@@ -386,7 +386,7 @@
         },
         {
           "created": "2010-01-23T04:56:22Z",
-          "lastUsed": "2010-01-23T04:56:22Z",
+          "last_used": "2010-01-23T04:56:22Z",
           "name": "ManagerName",
           "value": "Mark"
         }
@@ -407,7 +407,7 @@
       "value": [
         {
           "created": "2010-01-23T04:56:22Z",
-          "lastUsed": "2010-01-23T04:56:22Z",
+          "last_used": "2010-01-23T04:56:22Z",
           "name": "mozdef1.private.scl3.gmail",
           "value": "5a5munnfxYjqkaN0su1Kl7USxbqkILQN"
         }
--- a/docs/rfcs/UserProfilesv2.json
+++ b/docs/rfcs/UserProfilesv2.json
@@ -1,4 +1,4 @@
-﻿{
+{
   "_schema": "https://person-api.sso.mozilla.com/schema/v2/profile",
   "user_id": {
     "metadata": {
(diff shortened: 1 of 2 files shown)


docs/rfcs/UserProfilesv2.json
```diff
@@ -1,276 +1,435 @@
...
-	}
+  "_schema": "https://person-api.sso.mozilla.com/schema/v2/profile",
+  "user_id": {
+    "metadata": {
+      "publisher": "auth0",
+      "signatures": [
+        {
+          "name": "ZOWSLXKxx..",
+          "value": "LALALA...",
+            "type": "jwks"
```
gene1wood: nit : All the `type` field indentations are off (maybe tabs vs spaces)

docs/rfcs/UserProfilesv2_schema.json
```diff
@@ -218,130 +286,44 @@
                 "metadata",
```
gene1wood: Should this be instead
```
                 "metadata": { "$ref": "#/definitions/Metadata"},
```

docs/rfcs/UserProfilesv2_schema.json
```diff
@@ -218,130 +286,44 @@
...
                 "value"
```
gene1wood: And this
```
                 "vaue": { "$ref": "#/definitions/ValueElement"}
```

docs/rfcs/UserProfilesv2_schema.json
```diff
@@ -201,9 +269,9 @@
                 "metadata",
```
gene1wood: Should this be instead
```
                 "metadata": { "$ref": "#/definitions/Metadata"},
```

docs/rfcs/UserProfilesv2_schema.json
```diff
@@ -201,9 +269,9 @@
...
                 "value"
```
gene1wood: Should this be something that defines `value` as strictly a boolean?
```
  "value": {
    "type": "boolean"
  }
```

docs/rfcs/UserProfilesv2.json
```diff
@@ -1,276 +1,435 @@
...
+            "value": "LALALA...",
+            "type": "jwks"
+          }
+        ],
+        "classification": "PUBLIC"
+      },
+      "value": [
+        {
+          "created": "2010-01-23T04:56:22Z",
+          "lastUsed": "2010-01-23T04:56:22Z",
```
gene1wood: Should this be `last_used`

docs/rfcs/UserProfilesv2_schema.json
```diff
@@ -1,144 +1,206 @@
...
+        },
+        "ValueElement": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "created": {
+                    "type": "string",
+                    "format": "date-time"
+                },
+                "lastUsed": {
```
gene1wood: Should this be `last_used`

docs/rfcs/UserProfilesv2_schema.json
```diff
@@ -218,130 +286,44 @@
...
-                    "type": "string"
-                }
-            ],
-            "title": "ValueUnion"
+        "Name": {
+            "type": "string",
+            "title": "Name"
         },
-        "Signature": {
+        "Type": {
```
gene1wood: Is there any value in clarifying this schema type with instead of `Type` something like `SignatureType`?

docs/rfcs/UserProfilesv2.json
andrewkrug: maybe we should convert to YAML.  In theory the standard supports it.

docs/rfcs/UserProfilesv2_schema.json
andrewkrug: I second the disambiguation on Type here.  Since it's a reserved work in a lot of cases.  Better to be explicit.

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: no, refs are for "object types", this is for which attribute is required

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: no, refs are for "object types", this is for which attribute is required

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: no, refs are for "object types", this is for which attribute is required

docs/rfcs/UserProfilesv2_schema.json
gdestuynder: yes good catch

PR #112, opened 2018-06-27
Profilev2.999

gdestuynder:
Notes of importance:

- `last_modified` per sub-field is no longer present. are we ok with this? ex, if you verify an identity you only know the last time *any* has been modified but not a particular identity

- `verified` per sub-field is no longer present. This means we document the assumption of what's verified, vs storing it in the profile data. are we ok with this?

- this is a separate branch merge, not `master` so that CIS work for profile v2 does not affect `master` until we have CIS and personapi code ready for it

See the docs for additional info (rendered: https://github.com/gdestuynder/cis/blob/profilev2.999/docs/Profiles.md - click the sample yaml profile at the bottom for a TLDR - read the diff only after you read these!)

If you merge, please squash-merge (this is several commits of work but should be represented as one)


johngian:
After going through the current UX work it looks like we also need to add a field in order to allow user's to store their full name in local language.

docs/user_profile_core_plus_extended.yml
```diff
@@ -0,0 +1,558 @@
...
+      typ: 'JWT'
+      value: ''
+  metadata:
+    classification: PUBLIC
+    last_modified: '2018-01-01T00:00:00Z'
+    created: '2018-01-01T00:00:00Z'
+    publisher_authority: mozilliansorg
+  values:
+    - French
+      English (US)
```
fiji-flo: this should be a list:
```yaml
- French
- English (US)
- German
```

fiji-flo:
Does anyone remember what happened to `tshirt_size`?

gdestuynder:
@fiji-flo i think we just killed it at some point, deciding it wouldnt be in the profile. though it could go in extended if we use the profile data as 100% user profile backend (vs the split backends we have today between cis and mozilliansorg)

gdestuynder:
added `alternate_name` in the extended profile

docs/user_profile_core_plus_extended.yml
fiji-flo: No it's not :confused: (or am I missing something here?).

Quoting [yaml.org](http://yaml.org/spec/1.2/spec.html#id2797293):
> A block sequence is simply a series of nodes, each denoted by a leading “-” indicator.

It also renders wrong in the according [JSON](https://github.com/mozilla-iam/cis/pull/112/files#diff-63a3d9c31029bcc54e825c6c3fcbf858R543) which is generated from this YAML, isn't it?

johngian:
@gdestuynder Since we already have the HR first and last name should we just use a ``full_name`` field instead for the non-HRIS full name? It's pretty problematic to maintain first/last names because user's names might be more complicated (aka not everyone necessarily has a first and a last name).

gdestuynder:
> @gdestuynder Since we already have the HR first and last name should we just use a full_name field instead for the non-HRIS full name? It's pretty problematic to maintain first/last names because user's names might be more complicated (aka not everyone necessarily has a first and a last name).

No, we already went through this one: the issue with a full name is that you do not know how to split it back into 2 fields (first and last name), and most RPs (as well as OIDC) [uses separate fields](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims)

You can make the UX pass an empty name though, which should work fine.

docs/user_profile_core_plus_extended.yml
gdestuynder: the yaml is generated from the json, though i might have messed something up ill check that

gdestuynder: so yeah my bad, this is a list not a list of dict, fixing :)

gdestuynder:
just a reminder comment that this needs a squash-merge for prettyness, when clicking that github merge button

gdestuynder:
> Does anyone remember what happened to tshirt_size?
Following up on my previous question:
made an executive decision of not supporting it

Rationale: we ask t-shirt size again every time we *actually* need it, because people's size change over time, or they want tshirts for their friends, etc. Thus, the value is not actually very useful. I suspect it was our initial assessment, but we might have omitted to record it

johngian:
:rocket:

PR #113, opened 2018-07-10
well-known endpoint + sigs

gdestuynder:
spec for the well-known endpoint and jwks cis publisher + apps.yaml
signatures


gdestuynder:
hub cli program got me. wrong branch. reopening

PR #115, opened 2018-07-12
house keeping for profile v2 cis v2

andrewkrug:
I deleted most of what's here and will strategically be pulling in concepts we want to keep from cisv1.  

Mostly though this is blowing up the monolithic library into multiple libraries and abstracting business logic from the serverless functions.

Better interfaces.  More mocks.  Two faker libraries.


docs/Event.md
```diff
@@ -4,11 +4,11 @@ Or how to send an identity event that CIS will correctly validate and understand
 
 ## Publisher key and CIS stream access
```
gdestuynder: i suspect we'll have a few more passes at this doc but wfm for now

gdestuynder: (in particular we should just link to an example)

python-modules/README.md
```diff
@@ -0,0 +1,120 @@
+# Mozilla Change Integration Python Modules
+
+Mozilla IAM uses a variety of python modules to achieve integration
+across the stack.  In version 1 of the change integration service one library
+formerly known as `cis` held all this functionality.  In v2 of the service separation of concerns requires a more loosely coupled approach.
```
gdestuynder: would be good to write "why" instead

python-modules/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Modules in this directory.  
+
+**cis-common**
+
+Code that is common to all additional libs.  Configuration management, authorization layers, etc.
+
+_User Stories_
+
+As a publisher I must exchange a CLIENT_ID and CLIENT_SECRET with auth0 for a bearer token.  This grants the ability to publish changes.
```
gdestuynder: `access token`
i know the text in the header says bearer, but it's an access token =)

python-modules/README.md
gdestuynder: define "publisher" and "processor" here

docs/images/CIS-AWS-Stencils.xml
```diff
@@ -1 +1,2 @@
...
\ No newline at end of file
+<?xml version="1.0" encoding="UTF-8"?>
```
gdestuynder: is there a render of this anywhere?

andrewkrug: Yeah it's the DOC

gdestuynder:
added a commit with nits instead of comments to save time hopefully

PR #118, opened 2018-07-18
Fix schema key and make it required across different schema definitions.

johngian:


johngian:
@kang r?

gdestuynder:
we already define it here: https://github.com/johngian/cis/blob/28afe47d79812e7a647e54be59674427b4b619ce/docs/profile_data/profile.schema#L2

I think there's a confusion between the **profile json data** `schema` field and the **profile schema**'s `$schema` field (reserved, must be present). To my knowledge json-schema.org does not requires or dictate any content in the json data itself.

We added `schema` in the json data so that there is a pointer back to the schema - that said, since `$` is used to denotate **schema** keywords, I would think having it used in the json data as well would bring more confusion (since its not required or mentioned by json-schema.org in this case)

@johngian thoughts? Or, of course, I could also be missing something

PR #120, opened 2018-07-23
Add Docs and Cover pyconfig

andrewkrug:
* Add documentation
* Add support for a pyconfig file
* Change some of the abstraction of assumeRole information.


python-modules/cis_aws/setup.cfg
```diff
@@ -16,7 +16,7 @@ universal = 1
 
 [flake8]
 exclude = docs
-max-line-length=119
+max-line-length = 120
```
gdestuynder: wat ! ;-)

PR #127, opened 2018-08-28
Land crypto lib in profile v2

andrewkrug:
This completes the minimum viable product on the crypto lib.  

* Provides sign verify on the cli as well as objects and configuration options to verify in the CIS pipeline.


python-modules/cis_crypto/cis_crypto/operation.py
```diff
@@ -73,20 +82,63 @@ def _get_public_key(self):
                     )
                 )
             )
-            key_name = self.config('public_key_name', namespace='cis')
+            key_name = self.config('public_key_name', namespace='cis', default='access-file-key')
             file_name = '{}.pub.pem'.format(key_name)
             fh = open((os.path.join(key_dir, file_name)), 'rb')
```
gene1wood: This could be in a `with` block to close the file

python-modules/cis_crypto/cis_crypto/operation.py
```diff
@@ -73,20 +82,63 @@ def _get_public_key(self):
...
+        elif self.well_known_mode == 'http' or self.well_known_mode == 'https':
             # Go get it from the .well-known endpoint and load as json
             # return a dictionary of the json loaded data
-            pass
+            well_known_url = self.config('well_known_url', namespace='cis')
+            res = requests.get(well_known_url)
+
+            if res.status_code == 200:
+                key_list = self._reduce_keys(res.json())
+                return key_list
```
gene1wood: Maybe return `False` if the well-known fetch fails? This just returns nothing, not sure how your upstream code deals with this method returning nothing (maybe setting a variable to the return value of a method that doesn't return sets it to `None`, I'm note sure)

python-modules/cis_crypto/cis_crypto/operation.py
```diff
@@ -73,20 +82,63 @@ def _get_public_key(self):
...
+
+    def _reduce_keys(self, well_known_response):
+        access_file_keys = well_known_response['access_file']['jwks_keys']
+        publisher_keys = well_known_response['publishers_supported']
+
+        keys = []
+
+        for key in access_file_keys:
```
gene1wood: Wait, so why iterate over the content of `access_file_keys` if you don't use the content. Could you instead do
```
for i in range(len(access_file_keys)):
    keys.append(
        {
            'access_file_key_{}'.format(i): access_file_keys.get(str(i))
        }
    )
```

python-modules/cis_crypto/cis_crypto/operation.py
```diff
@@ -73,20 +82,63 @@ def _get_public_key(self):
...
+            count = 0
+            keys.append(
+                {
+                    'access_file_key_{}'.format(count): access_file_keys.get(str(count))
+                }
+            )
+            count = count + 1
+
+        for k in publisher_keys:
+            count = 0
```
gene1wood: Don't need `count` here, we're not using it.

python-modules/cis_crypto/cis_crypto/operation.py
```diff
@@ -73,20 +82,63 @@ def _get_public_key(self):
...
         """Assumes you loaded a payload.  Return the same jws or raise a custom exception."""
-        jwk = self._get_public_key()
-        sig = jws.verify(self.jws_signature, jwk, algorithms='RS256', verify=True)
-        # XXX TBD failed verification will raise JWSError log in mozilla-iam format the custom exception
-        return sig
+        key_material = self._get_public_key()
+        if isinstance(key_material, dict):
+            sig = jws.verify(self.jws_signature, key_material, algorithms='RS256', verify=True)
+
+        if isinstance(key_material, list):
```
gene1wood: `elif` instead of `if`, easier to follow

python-modules/cis_crypto/cis_crypto/operation.py
```diff
@@ -73,20 +82,63 @@ def _get_public_key(self):
...
+                for k_id in key_material[key_material.index(key)]:
+                    logger.info('Attempting to match against: {}'.format(k_id))
+                    key_content = key_material[key_material.index(key)][k_id]
+                    try:
+                        sig = jws.verify(self.jws_signature, key_content, algorithms='RS256', verify=True)
+                        logger.info('Matched a verified signature for: {}'.format(k_id))
+                        return sig
+                    except JWSError as e:
+                        logger.error(e)
+        raise JWSError('The signature could not be verified for any trusted key.')
```
gene1wood: Maybe instead set `sig` to `False` and have your upstream code check if `sig` is not False (instead of raising an exception), just an idea, this way looks fine too.

python-modules/cis_crypto/cis_crypto/operation.py
```diff
@@ -18,8 +24,11 @@ def __init__(self):
 
     def load(self, data):
```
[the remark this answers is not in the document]
gene1wood: And it looks like you don't use the return value in the upstream code.

python-modules/cis_crypto/cis_crypto/cli.py
```diff
@@ -53,7 +45,36 @@ def parse_args(self, args):
             help='The path to the file you would like to sign.'
         )
 
+        verify_operation_parser.set_defaults(func='verify_operation')
         return parser.parse_args(args)
 
     def run(self):
+        logger = logging.getLogger('cis_crypto')
```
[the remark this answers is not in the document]
gene1wood: Could you use logging propagation to do this? `propagate` is True by default

python-modules/cis_crypto/cis_crypto/cli.py
```diff
@@ -53,7 +45,36 @@ def parse_args(self, args):
...
+            )
+            file_content = common.load_file(self.config.file)
+            o = operation.Verify()
+            o.load(file_content)
+            try:
+                jws = o.jws()  # This will raise if the signature is invalid.
+                logger.info('Signature verified for file: {}'.format(self.config.file))
+            except jose.exceptions.JWSError:
+                logger.error('The signature could not be verified.')
+                sys.exit(-1)
```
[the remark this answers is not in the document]
gene1wood: No, the default "something went wrong" exit code in posix is 1

python-modules/cis_crypto/cis_crypto/cli.py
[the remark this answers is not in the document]
andrewkrug: Tell me more!

python-modules/cis_crypto/cis_crypto/cli.py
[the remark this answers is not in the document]
andrewkrug: I changed this.

python-modules/cis_crypto/cis_crypto/operation.py
```diff
@@ -28,6 +28,8 @@ def load(self, data):
             data = yaml.safe_load(data)
         except yaml.scanner.ScannerError:
             logger.info('This file is likely not YAML.  Attempting JSON load.')
+        except AttributeError:
```
gene1wood: "[An except clause may name multiple exceptions as a parenthesized tuple](https://docs.python.org/3/tutorial/errors.html#handling-exceptions)"
```
         except (yaml.scanner.ScannerError, AttributeError): 
             logger.info('This file is likely not YAML.  Attempting JSON load.') 
```

PR #131, opened 2018-09-13
s/JWT/JWS/

gdestuynder:
We now use JWS instead of JWT so this fixes the doc and more importantly all schemas, since signing with JWS would otherwise fail validation


gdestuynder:
wrong branch kang

PR #138, opened 2018-09-18
Profilev2

andrewkrug:


python-modules/cis_aws/cis_aws/connect.py
```diff
@@ -156,8 +153,10 @@ def input_stream_client(self):
     def _check_sessions_exist(self):
         if self._discover_cis_environment() == 'local':
             logger.info('CIS Local environment detected skipping cloud based validations.')
+            return
```
gdestuynder: why not `elif`?

python-modules/cis_change_service/cis_change_service/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+__version__ = '0.0.1'
+
+import os
+
+from everett.manager import ConfigManager
+from everett.manager import ConfigIniEnv
+from everett.manager import ConfigOSEnv
+
+
+def get_config():
```
gdestuynder: i think its missing `__all__ = [get_config]

python-modules/cis_change_service/cis_change_service/api.py
```diff
@@ -0,0 +1,83 @@
+import logging
```
gdestuynder: careful with the scope, thats a lot of defines ;-)

python-modules/cis_change_service/cis_change_service/api.py
```diff
@@ -0,0 +1,83 @@
...
+@app.errorhandler(AuthError)
+def handle_auth_error(ex):
+    response = jsonify(ex.error)
+    response.status_code = ex.status_code
+    return response
+
+
+@app.route('/')
+def index():
+    return 'Mozilla Change Integration Service Endpoint'
```
gdestuynder: not 403?

python-modules/cis_change_service/cis_change_service/api.py
```diff
@@ -0,0 +1,83 @@
...
+    user_profile = request.get_json(silent=True)
+    publish = operation.Publish()
+    result = publish.to_stream(user_profile)
+
+    if config('stream_bypass', namespace='cis', default='false') == 'true':
+        # Plan on stream integration not working an attempt a write directly to discoverable dynamo.
+        # Great for development, seeding the vault, and contingency.
+        logger.debug(
+            'Stream bypass activated.  Integrating user profile directly to dynamodb for: {}'.format(
+                user_profile.get('user_id').get('value')
```
gdestuynder: might want to use DotDicts here and elsewhere (Florian also uses them) so that you can write:
`user_profile.user_id.value`

if using `cis_profile` it supports this by default as well, which means you could do:

```
import cis_profile
...

def change():
    user_profile = cis_profile.User(request.get_json(silent=True))
    ....
    result = publisher.to_stream(user_profile.as_json())
    ...
    logger.debug(...user_profile.user_id.value)
```
etc

python-modules/cis_change_service/cis_change_service/idp.py
```diff
@@ -0,0 +1,131 @@
...
+
+from cis_change_service import get_config
+from cis_change_service.exceptions import AuthError
+
+logger = logging.getLogger(__name__)
+
+CONFIG = get_config()
+
+AUTH0_DOMAIN = CONFIG('auth0_domain', namespace='change_service', default='auth-dev.mozilla.auth0.com')
+API_IDENTIFIER = CONFIG('api_identifier', namespace='change_service', default='https://change.sso.allizom.org')
```
gdestuynder: maybe this should be api.sso.allizom.org ;-)

python-modules/cis_change_service/cis_change_service/idp.py
```diff
@@ -0,0 +1,131 @@
...
+    jwks = json.loads(jsonurl.read())
+    return jwks
+
+
+def requires_auth(f):
+    """Determines if the Access Token is valid
+    """
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        if CONFIG(
```
gdestuynder: hard to read

python-modules/cis_change_service/cis_change_service/profile.py
```diff
@@ -0,0 +1,97 @@
...
+        if self.identity_vault_client is None:
+            self._connect()
+
+        user_id = self._get_id(profile_json)
+        primary_email = self._get_primary_email(profile_json)
+        dynamodb_schema_dict = {
+            'profile': {'S': json.dumps(profile_json)},
+            'id': {'S': user_id},
+            'sequence_number': {'S': self.sequence_number},
+            'primaryEmail': {'S': primary_email}
```
gdestuynder: i wonder if you want to change this with profilev2 (it doesnt have `primaryEmail` as a field) (i know this means a dynamodb change)

python-modules/cis_change_service/cis_change_service/profile.py
```diff
@@ -0,0 +1,97 @@
...
+        self.connection_object.assume_role()
+        self.identity_vault_client = self.connection_object.identity_vault_client()
+        return self.identity_vault_client
+
+    def put_profile(self, profile_json):
+        """Write profile to the identity vault."""
+        if self.identity_vault_client is None:
+            self._connect()
+
+        user_id = self._get_id(profile_json)
```
gdestuynder: i wonder if the vault should use `cis_profile` here

python-modules/cis_change_service/tests/test_api.py
```diff
@@ -0,0 +1,266 @@
...
+            },
+            data=json.dumps(user_profile),
+            content_type='application/json',
+            follow_redirects=True
+        )
+
+        assert result.status_code == 200
+
+    def teardown_class(self):
+        subprocess.Popen(['killall', 'node'])
```
gdestuynder: i would rather not do that since it kills all node processes.
instead we should trap and kill childs, or record and kill pids

theres fanciers ways (including trapping a signal), but heres a straigthforward way (that does not work on Windows, but works everywhere else):
```
self.dynaliteprocess =subprocess.Popen("dynalite, etc", preexec_fn=os.setsid)
....
def __delete__(self):
    os.killpg(os.getpgid(self.dynaliteprocess.pid), 15)

```

gdestuynder: same for other files

gdestuynder:
free unsollicited review ;-)

python-modules/cis_change_service/cis_change_service/idp.py
andrewkrug: these are going to end up being different backend flask apps and I'm not sure at present how to unify them.  So change in my mind was separate from api.

python-modules/cis_change_service/cis_change_service/idp.py
andrewkrug: Do you mean because it's a decorator?  Or ...

python-modules/cis_change_service/cis_change_service/profile.py
andrewkrug: There is a value I can pull out that is the users email... that's what is happening at present.

python-modules/cis_change_service/cis_change_service/profile.py
andrewkrug: It should.  I suggest backlogging integrating cis_profile to after we finish the processor modules.

python-modules/cis_change_service/tests/test_api.py
andrewkrug: This is great.

python-modules/cis_change_service/cis_change_service/api.py
andrewkrug: cis_profile did did exist when I authored this.  Also ... I believe that user profile.validate() does raise as is and interrupts ... but this makes it easier to understand.

python-modules/cis_change_service/cis_change_service/api.py
andrewkrug: For / it's kinda nice to have something that returns the version and status code 200 so we can health check it.

python-modules/cis_aws/cis_aws/connect.py
andrewkrug: where would the `elif` go?

python-modules/cis_change_service/cis_change_service/api.py
andrewkrug: All the logging is put in haphazard (like this) at present because it'll be replaced with cis_logging at the end of the build.

python-modules/cis_aws/cis_aws/connect.py
andrewkrug: ah I see... never-mind.  I'll change this.  But would say this is a style choice.

python-modules/cis_change_service/cis_change_service/profile.py
andrewkrug: What should I put here instead? `'primaryEmail': {'S': primary_email}`

andrewkrug: We have to store primary email as a top level attribute separate from the map because it needs to be searchable in a global secondary index.  I may be misinterpreting this... looking at the profile.  Do you mean s/primaryEmail/primary_email/g ?

python-modules/cis_change_service/cis_change_service/__init__.py
andrewkrug: I need to go back and do this everywhere...

python-modules/cis_change_service/cis_change_service/profile.py
andrewkrug: If so this is an easy change.  I also need to update the schema to add a partiton key to this.

python-modules/cis_change_service/cis_change_service/api.py
gdestuynder: i meant more that style wise you often have a lot of "from x import z" and end up with a lot of "z", which generally makes it more likely to mistakenly override variables (specially in the future). style choice

python-modules/cis_change_service/cis_change_service/api.py
gdestuynder: good point

python-modules/cis_change_service/cis_change_service/api.py
gdestuynder: yeah it does raise - i though about this a bit while writing it, and while i liked the idea of raising, it could also just return a valid or invalid result (only ldap2s3 uses it atm so it can still be changed, otherwise def validate_with_return ;-)

python-modules/cis_change_service/cis_change_service/idp.py
gdestuynder: change.api.sso.allizom.org?
not a big deal either way

python-modules/cis_change_service/cis_change_service/idp.py
gdestuynder: no, its the syntax that is a little odd compared to the rest of the file because it breaks line several times with comparisons (`==`) in the middle, wasn't used to that
ie style stuff

python-modules/cis_change_service/cis_change_service/profile.py
gdestuynder: yes mainly s/primaryEmail/primary_email/ so it reflects the profile data field name

python-modules/cis_change_service/tests/test_api.py
gdestuynder: i forgot to note that this part is the reason i blocked on the review, the rest are style, comments, or just "hey did you think of that"

python-modules/cis_change_service/cis_change_service/api.py
andrewkrug: Actually this _is_ the behavior.  But it's being handled in the operation class.  Catching the exception down there and raising a 400 back up to make sure we continue when processing a batch.  See L#61 in cis_publisher/operation.py

python-modules/cis_change_service/cis_change_service/idp.py
andrewkrug: You're totally right this was hard to reason about.  I fixed it.

python-modules/cis_change_service/cis_change_service/profile.py
andrewkrug: Added a note to come back into this later.

python-modules/cis_change_service/cis_change_service/api.py
gdestuynder: ok :)

python-modules/cis_aws/tests/test_connect.py
```diff
@@ -463,4 +463,5 @@ def test_pyconfig_support(self):
         assert test_port == '34567'
 
     def teardown(self):
```
gdestuynder: just a note, i think `__delete__` would be the pythonic way (it also trigger on object being deleted automatically)

PR #139, opened 2018-09-18
ensure we always load schema no matter what (its not needed if you don't     validated, but why wouldn't you?) allow for users of the class to retrieve     the cached schema

gdestuynder:


gdestuynder:
nope, hub-pullrequest doesnt know how to make branch req

PR #142, opened 2018-09-21
some quick fixes to facilitate cis_crypto deployment in non-aws envs

gdestuynder:


python-modules/cis_crypto/README.md
```diff
@@ -47,14 +47,16 @@ Cryptography Settings for sign-verify operations
...
+cis_public_key_name=fake-access-file-key-public.pem # Optional for use with file mode only.
+cis_signing_key_name=access-file-key-private.pem # Not optional!
 
 ## AWS Specific Secret Manager Settings
 secret_manager_ssm_path=/iam
 secret_manager_ssm_region=us-west-2
 
-# give the name of the key without file extension that you would like to sign assertions with
-signing_key_name=access-file-key
+## File Specific Secret Manager Settings (to use with `secret_manger=file`)
```
andrewkrug: nit... can we just call it "secret_manager_file_path". It's already in the CIS namespace.  That would mean if you want to use OSEnv provider that the variable would be CIS_CIS_SECRET_MANAGER_FILE_PATH

gdestuynder: right now it uses the `cis` namespace so the README doesnt work hence the fix, else, would have to change the code for this too.
i see i left a typo though (manger)

PR #145, opened 2018-09-26
Remove TAI (no longer wanted), fix jwks, ...

gdestuynder:
Fix jwks specification to follow
https://tools.ietf.org/html/rfc7517#section-5 correctly
Drop unused SHA1 thumb parameter from jwks


gdestuynder:
note: the RFC spec makes it so that some of the code you're working on needs to use the new structure :(

gdestuynder:
new commit move the signature field around as chatted about

PR #147, opened 2018-09-27
Profilev2

andrewkrug:
* WIP on schema changes
* vault module complete


python-modules/cis_crypto/cis_crypto/operation.py
```diff
@@ -97,42 +99,37 @@ def _get_public_key(self):
...
-                for k_id in key_material[key_material.index(key)]:
-                    logger.info('Attempting to match against: {}'.format(k_id))
-                    key_content = key_material[key_material.index(key)][k_id]
-                    try:
-                        sig = jws.verify(self.jws_signature, key_content, algorithms='RS256', verify=True)
-                        logger.info('Matched a verified signature for: {}'.format(k_id))
-                        return sig
-                    except JWSError as e:
-                        logger.error(e)
+                key.pop('x5t', None)
```
gdestuynder: note i dont think we care about `x5t`. its fine to leave it in there, but i dropped it from the schema because its not very useful
x5t being the thumbprint (a sha1 hash of the cert)
x5c being the actual cert (which you can do a .. sha256 hash of.. or sha1)

PR #151, opened 2018-09-28
[READY-TO-MERGE] 20180928 pv2 [SQUASHME-IF-MERGED]

gdestuynder:
- [x] refactor load_schema into class
- [x] load well-known from cache as well (not just schema)
- [x] support TTL in cache (default 3600s)
- [x] add classification support
- [x] support a new graphene set of classes for user ObjectTypes


gdestuynder:
@andrewkrug my understanding is that i cannot access the data in the `graphene.ObjectType` class scope to understand if the fields have a certain classification set.
I can only return "unauthorized" or similar if the graphene context tells me that you dont have access already, or, by creating `ObjectType`s which have a different set of classifications (but since these may change dynamically it wouldnt really work)

that would mean we have this set in the authorization middleware which lookups the results.

note that similarly graphene can do a lot of validation, but we already do this in schema validation. the only sane way to add the graphene validation (for convenience) seems to me that the profile generator is the same code as the graphql generator and that's quite a bit of extra work if we did that

gdestuynder:
we decided the authorization layer will be done in another module which will call cis_profile's future `User.filter_scope(xxx)`

PR #152, opened 2018-10-02
20181001 pv2 scope

gdestuynder:
`User.filter_scopes(MozillaDataClassification.PUBLIC)`
and then User has only public scopes


gdestuynder:
squashmerge + rebase seems to be really not nice in git - re-pr'ing with a fresh base

PR #167, opened 2018-11-02
Profilev2 display and other changes

gdestuynder:
See commit msg ;-)
Note:
- still have to port over validator
- still have to fix the code to work with this
- it actually does validate, even thus the makefile doesnt support validating it yet
- supersede work done in https://github.com/mozilla-iam/cis/pull/163


python-modules/cis_profile/cis_profile/profile.schema
```diff
@@ -1,109 +1,173 @@
...
+                    "$ref": "#/definitions/StandardAttributeValues",
+                    "$comment": "The languages the user would like to use to communicate with others."
                 },
                 "tags": {
-                    "$ref": "#/definitions/StandardAttributeValues"
-                },
-                "pronouns": {
-                    "$ref": "#/definitions/StandardAttributeString"
+                    "$ref": "#/definitions/StandardAttributeValues",
+                    "$comment": "Just tags. Used for Mozillians.org style vouches, for example."
```
akatsoulas: This can be a bit misleading. Regardless of vouches, which may be an access group, I would describe this something like a union of interests. Eg python tag is a self descriptive tag for people with a common interest in python

python-modules/cis_profile/cis_profile/user_profile_null.json
```diff
@@ -146,7 +151,9 @@
       "classification": "MOZILLA CONFIDENTIAL",
       "last_modified": "1970-01-01T00:00:00Z",
       "created": "1970-01-01T00:00:00Z",
-      "verified": false
+      "verified" : true,
+        "display": null,
```
akatsoulas: Indentation is a bit off here.

python-modules/cis_profile/cis_profile/user_profile_null.json
```diff
@@ -171,7 +178,9 @@
       "classification": "PUBLIC",
       "last_modified": "1970-01-01T00:00:00Z",
       "created": "1970-01-01T00:00:00Z",
-      "verified": false
+      "verified" : true,
+        "display": null,
```
akatsoulas: Same here.

python-modules/cis_profile/cis_profile/user_profile_null.json
```diff
@@ -196,7 +205,8 @@
       "classification": "PUBLIC",
       "last_modified": "1970-01-01T00:00:00Z",
       "created": "1970-01-01T00:00:00Z",
-      "verified": false
+      "verified" : true,
+        "display": null
```
akatsoulas: And here.

akatsoulas: This has a few more other occurrences.

akatsoulas:
Will the docs be updated in a different PR or will be dropped completely?

andrewkrug:
On principal I'm fine with the new structure.  No need to update comprehensive docs but I would ask that we update the sample profile json and yaml files as well as the null files in this commit.

andrewkrug:
In order to test all of our integration we will also need to update the profile_faker and the ldap2s3 code to match this.

gdestuynder:
@akatsoulas docs will be updated in a different PR, just making sure we agree on the current structure before i rewrite a bunch of things (it seems we do though, yay!)

gdestuynder:
> In order to test all of our integration we will also need to update the profile_faker and the ldap2s3 code to match this.

note that the profile_faker does not support the current (pre-PR) profile either

gdestuynder:
it is done.
@andrewkrug 

Note: particular attention is needed on the new `profile.schema`.

PR #168, opened 2018-11-02
Fix top level tests

andrewkrug:


python-modules/Makefile
```diff
@@ -11,6 +11,9 @@ build: $(MODULES)
 test: tests
 tests: $(MODULES)
 
+test-tox: test-tox
```
gdestuynder: shouldnt we just use the same test targets for all?

PR #188, opened 2018-11-15
 Add primary email for identities.

fiji-flo:


python-modules/cis_profile/cis_profile/data/profile.schema
```diff
@@ -625,6 +626,7 @@
             },
             "mozilliansorg_id": { "$ref": "#/definitions/StandardAttributeString" },
             "bugzilla_mozilla_org_id": { "$ref": "#/definitions/StandardAttributeString" },
+            "bugzilla_mozilla_org_primary_email": { "$ref": "#/definitions/StandardAttributeString" },
             "mozilla_ldap_id": {
```
gdestuynder: missing `mozilla_ldap_primary_email` 
it should have the same attributes as `mozilla_ldap_id` of course

PR #191, opened 2018-12-13
productionalize WIP on deploying dev cisv2

andrewkrug:


serverless-functions/Makefile
```diff
@@ -0,0 +1,57 @@
...
+.PHONY:clean-layer
+clean-layer:
+	rm -rf layer/*
+	rm -rf build/*
+	rm cis.zip
+	rm serverless-functions/cis.zip
+
+.PHONY:package-layer
+package-layer:
+	mkdir -p $(PY_DIR)
```
gdestuynder: probably want to add PY_DIR to the clean target, as well as have the clean target call clean-layer (so clean cleans all)

note: you can `-rm xx` or `rm -f` to ensure make wont choke if the files dont exist, this is handy for clean since you just want it to keep running even if half of the things were already cleaned

serverless-functions/change/serverless.yml
```diff
@@ -0,0 +1,39 @@
+service: change-service
```
gdestuynder: sometimes i feel like we need to template the template hehe. maybe we'll end up with a readme of where all the vars are?

PR #218, opened 2019-01-15
Add `uuid` field support which is opaque and managed by CIS

### LINKED ISSUE #217: add a `uuid` field in profilev2

gdestuynder:
yup
fixes #217 

note also fixed the stray preference field

### DIFF
--- a/python-modules/cis_profile/cis_profile/data/profile.schema
+++ b/python-modules/cis_profile/cis_profile/data/profile.schema
@@ -12,7 +12,7 @@
                 "uuid": {
                   "allOf": [
                       { "$ref": "#/definitions/StandardAttributeString" },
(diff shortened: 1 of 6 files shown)


python-modules/cis_profile/cis_profile/data/profile.schema
```diff
@@ -9,6 +9,14 @@
                 "schema": {
                     "type": "string"
                 },
+                "uuid": {
+                  "allOf": [
+                      { "$ref": "#/definitions/StandardAttributeString" },
+                      { "$ref": "#/definitions/DisplayAnyValue" },
```
fiji-flo: I make this `DisplayPublicOnly` since it might be displayed in the url.

PR #223, opened 2019-01-16
Support transactions -- multiple fixes

andrewkrug:


python-modules/cis_change_service/cis_change_service/__init__.py
```diff
@@ -4,9 +4,10 @@
 
 __version__ = '0.0.1'
```
gdestuynder: remove
```suggestion

```

python-modules/cis_change_service/cis_change_service/profile.py
```diff
@@ -15,80 +21,105 @@ def __init__(self, sequence_number=None):
...
         self.identity_vault_client = self.connection_object.identity_vault_client()
         return self.identity_vault_client
 
+    def _verify(self, profile_json):
+        cis_profile = User(user_structure_json=profile_json)
+        try:
+            if self.config('verify_publishers', namespace='cis') == 'true':
+                cis_profile.verify_all_publishers()
+
+            if self.config('verify_signatures', namespace='cis') == 'true':
```
gdestuynder: bool comp as string! ;-)

PR #227, opened 2019-01-16
basic version of a parser that will generate CIS profiles instead of

gdestuynder:
storing the raw HRIS data to S3
This allow for immediately disgarding any unused data instead of having
it stored, and also makes it easier to submit to the CIS /change API


gdestuynder:
this "Saves" the v1 out of the way

PR #231, opened 2019-01-17
custom_x_primary_email+dinopark

gdestuynder:


python-modules/cis_profile/cis_profile/fake_profile.py
```diff
@@ -86,7 +86,7 @@ def fxa():
             #    "custom_1": custom,
```
fiji-flo: If we uncomment this we have it in fake profiles.

PR #288, opened 2019-01-18
Add support for display filtering

fiji-flo:


python-modules/cis_profile/cis_profile/profile.py
```diff
@@ -486,3 +482,18 @@ def _sign_attribute(self, attr, publisher_name):
         sigattr["typ"] = "JWS"  # ""
         sigattr["value"] = self.__signop.jws()
         return attr
+
+
+def filter(level, valid, check):
```
gdestuynder: please use another name, filter is a reserved keyword

PR #307, opened 2019-01-22
improve test run times

andrewkrug:


gdestuynder:
```
E           botocore.errorfactory.ResourceInUseException: An error occurred (ResourceInUseException) when calling the CreateStream operation: Stream local-stream under account 000000000000 already exists.
../.cis-env/cis_publisher/lib/python3.7/site-packages/botocore/client.py:661: ResourceInUseException
```
maybe we could/should start dyna/kine once for all tests and run a cleanup in between each module
(or just increment the port by 1 for each test)

PR #309, opened 2019-01-22
Test improvements

andrewkrug:


gdestuynder:
note: random might sometimes fail

PR #312, opened 2019-01-24
Top level username.

### LINKED ISSUE #304: add a `primary_username` as primary, indexed field. also allow it to be queriably by the api

fiji-flo:
resolves #304

### DIFF
--- a/python-modules/cis_profile/cis_profile/fake_profile.py
+++ b/python-modules/cis_profile/cis_profile/fake_profile.py
@@ -312,6 +312,7 @@ class FakeUser(cis_profile.profile.User):
         identities, user_id, email = fake.ldap_identity()
 
         self._d("user_id.value", user_id)
+        self._d("primary_username.value", "r--{}".format(base64.urlsafe_b64decode(fake.uuid4())))
         self._d("login_method.value", "Mozilla-LDAP")
         self._d("active.value", fake.boolean(chance_of_getting_true=config._active_percent))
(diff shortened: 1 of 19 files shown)


python-modules/cis_profile/cis_profile/fake_profile.py
```diff
@@ -161,11 +161,16 @@ def login_method(self):
...
         return self.generator.random.choice(p)
 
-    def usernames(self, additional=True):
-        u = {"mozilliansorg": self.generator.user_name()}
-        if additional:
-            for _ in range(self.generator.random.randrange(0, 10)):
-                u[self.generator.words(nb=1)[0]] = self.generator.user_name()
+    def primary_username(self):
+        if self.generator.boolean():
+            return "r--{}".format(base64.urlsafe_b64decode(fake.uuid4()))
```
gdestuynder: is this meant to be `r--`?

fiji-flo: Yes. Maybe I should add a comment?
This is how the random usernames look like in DinoPark these days. The `r--` prefix is used to determine that this is a random username.

I can leave it out here. But it's something we need to figure out otherwise.

The use case is that our front-end wants to hide _random_ usernames in the main profile card. And I thought that's a rather sustainable solution.

gdestuynder: ah ok, yeah a comment wouldnt be bad

PR #314, opened 2019-01-26
add signing and updating last modified timestamp

andrewkrug:


andrewkrug:
Let's wait to merge this until I add the corresponding IAM policies and secrets.  Else this will break the testing environment.

PR #315, opened 2019-01-26
Support User.merge

gdestuynder:
Ex:
```
        u_orig = User()
        u_patch = User()
        u_patch.timezone.value = "GMT -07:00 America/Los Angeles"
        u_orig.merge(u_patch, publisher="hris")
        print(u_orig.timezone.value)
        # >>> GMT -07:00 America/Los Angeles
```

### DIFF
--- a/python-modules/cis_profile/cis_profile/profile.py
+++ b/python-modules/cis_profile/cis_profile/profile.py
@@ -153,9 +153,12 @@ class User(object):
             try:
                 # current_attr is the original user attr
                 # level is the patch/merged in user attr
-                current_attr[_]["value"] = level[_]["value"]
+                # We skip null/None attributes even if the original/current user does not match (ie is not null)
+                if level[_]["value"] is not None:
+                    current_attr[_]["value"] = level[_]["value"]
             except KeyError:
-                current_attr[_]["values"] = level[_]["values"]
+                if level[_]["values"] is not None:
+                    current_attr[_]["values"] = level[_]["values"]
 
     def initialize_timestamps(self):
         now = self._get_current_utc_time()
--- a/python-modules/cis_profile/cis_profile/profile.py
+++ b/python-modules/cis_profile/cis_profile/profile.py
@@ -131,6 +131,8 @@ class User(object):
         # Default to top level
         if level is None:
             level = user_to_merge_in.__dict__
+        if _internal_level is None:
+            _internal_level = self.__dict__
 
         for attr in level.keys():
             # If this is an internal class object, skip it
@@ -138,25 +140,21 @@ class User(object):
                 continue
             # If we have no signature (or metadata in theory), this is not a "User attribute", keep doing deeper
             if "signature" not in level[attr].keys():
-                self.merge(user_to_merge_in, publisher=publisher, level=level[attr], _internal_level=attr)
+                self.merge(
+                    user_to_merge_in, publisher=publisher, level=level[attr], _internal_level=_internal_level[attr]
+                )
             # Check if this is an attribute we want to merge back
             elif level[attr]["signature"]["publisher"]["name"] == publisher:
                 tomerge.append(attr)
 
         for _ in tomerge:
             logger.debug("Merging in attribute {} for publisher {}".format(_, publisher))
-            print("==>", _, _internal_level)
-            current_attr = self.__dict__
-            if _internal_level is not None:
-                current_attr = current_attr[_internal_level]
 
-            # current_attr is the original user attr
+            # _internal_level is the original user attr
             # level is the patch/merged in user attr
             # We skip null/None attributes even if the original/current user does not match (ie is not null)
-            if level[_].get("value") is not None:
-                current_attr[_]["value"] = level[_]["value"]
-            elif level[_].get("values") is not None:
-                current_attr[_]["values"] = level[_]["values"]
+            if level[_].get("value") is not None or level[_].get("values") is not None:
+                _internal_level[_] = level[_]
 
     def initialize_timestamps(self):
         now = self._get_current_utc_time()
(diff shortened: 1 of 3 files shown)


python-modules/cis_profile/cis_profile/profile.py
```diff
@@ -100,6 +100,66 @@ def get_profile_from_file(self, user_structure_json_path):
...
+            # If this is an internal class object, skip it
+            if attr.startswith("_") or not isinstance(level[attr], dict):
+                continue
+            # If we have no signature (or metadata in theory), this is not a "User attribute", keep doing deeper
+            if "signature" not in level[attr].keys():
+                self.merge(user_to_merge_in, publisher=publisher, level=level[attr], _internal_level=attr)
+            # Check if this is an attribute we want to merge back
+            elif level[attr]["signature"]["publisher"]["name"] == publisher:
+                tomerge.append(attr)
+
```
fiji-flo:
```python
        for attr in tomerge:
            logger.debug("Merging in attribute {} for publisher {}".format(attr, publisher))

            try:
                # level[attr] is the patch/merged in current user
                # _intenal_level[attr] is in original user (self)
                # We skip null/None attributes even if the original/current user does not match (ie is not null)
                if level[attr].get("value") is not None or level[attr].get("values") is not None:
                     # copy everything including metadata and signatures
                    _intenal_level[attr] = level[attr]
```

python-modules/cis_profile/cis_profile/profile.py
```diff
@@ -100,6 +100,66 @@ def get_profile_from_file(self, user_structure_json_path):
...
+        attributes). The level MUST be from user_to_merge_in, not from the original/current user.
+        @_internal_level str attribute name  of previous level attribute from the current user. Used internally.
+
+        This function always returns None and is always successful.
+        """
+        tomerge = []
+        # Default to top level
+        if level is None:
+            level = user_to_merge_in.__dict__
+
```
fiji-flo:
```python
if _intenal_level is None:
    _inteanl_level = self.__dict__
```

python-modules/cis_profile/cis_profile/profile.py
```diff
@@ -100,6 +100,66 @@ def get_profile_from_file(self, user_structure_json_path):
...
+        for attr in level.keys():
+            # If this is an internal class object, skip it
+            if attr.startswith("_") or not isinstance(level[attr], dict):
+                continue
+            # If we have no signature (or metadata in theory), this is not a "User attribute", keep doing deeper
+            if "signature" not in level[attr].keys():
+                self.merge(user_to_merge_in, publisher=publisher, level=level[attr], _internal_level=attr)
```
fiji-flo:
```python
                 self.merge(user_to_merge_in, publisher=publisher, level=level[attr], _internal_level=_internal_level[attr]) 
```

python-modules/cis_profile/cis_profile/profile.py
```diff
@@ -100,6 +100,66 @@ def get_profile_from_file(self, user_structure_json_path):
             path = user_structure_json_path
         return DotDict(json.load(open(path)))
 
+    def merge(self, user_to_merge_in, publisher, level=None, _internal_level=None):
```
fiji-flo: I added some comments. I would carry on the `_internal level` the same way as `level`. Probably give them better names like `original_user_level` and `to_merge_in_user_level` or something.

Bigger question, do we need to merge everything from an attribute, including signatures and metadata?

I like how you deal with _null_ updated and I would add a flag to change this on demand (to enable field deletion).

python-modules/cis_profile/cis_profile/profile.py
```diff
@@ -483,17 +543,22 @@ def _sign_attribute(self, attr, publisher_name):
...
-            continue
-        if "metadata" not in level[attr].keys():
-            filter_all(valid=valid, level=level[attr], check=check)
-        elif level[attr]["metadata"][check] not in valid:
-            todel.append(attr)
-
-    for d in todel:
-        logger.debug("Removing attribute {} because it's not in {}".format(attr, valid))
-        del level[d]
+    def _filter_all(self, level, valid, check):
```
fiji-flo: See #316 
Sorry, I made the PR before looking at this.

python-modules/cis_profile/cis_profile/profile.py
gdestuynder: ok thats prettier. not in this case id drop the `try`

python-modules/cis_profile/cis_profile/profile.py
gdestuynder: @fiji-flo can you write some words about why this is needed?

gdestuynder: (or is that just so that `level` and `_internal_level` are both handled the same way? I don't think that changes much though, right?)

fiji-flo: Wow 2 typos in 2 lines :P

```python
if _internal_level is None:
    _internal_level = self.__dict__
```
would do the initialization of `_internal_level` when first called on the top level object. In all recursions `_internal_level` will be set already.
So yes this is to handle `level` and `_internal_level` the same way. What changes is: it replaces  [line 149](https://github.com/mozilla-iam/cis/pull/315/files#diff-6c6103298575e3cf2052ba6a215d0060R149):
and would allow merging of fields deeper than 2 levels.

This only works with my other comment in place.

python-modules/cis_profile/cis_profile/profile.py
fiji-flo: My thinking was if we plan to use this for partial updates via the change API. The updated fields would be signed by the RP (DinoPark most likely) already.
Or is this only meant to be used by an RP as a lib?

python-modules/cis_profile/cis_profile/profile.py
gdestuynder: side note: we never have more than 2 level

fiji-flo: Yes :P
But I would vote for either doing this iterative and only support 2 levels or recursive and support >2 levels.

PR #328, opened 2019-01-31
Fix e2e tests

andrewkrug:
Fixup e2e helper lib to include a sanitizing method for bad publishers.
Sign using the dev keys for the dev environment.

### DIFF
--- a/e2e/helpers.py
+++ b/e2e/helpers.py
@@ -1,0 +1,47 @@
+import boto3
+import os
+
+
+def get_cis_environment():
+    return os.getenv("CIS_ENVIRONMENT", "development")
+
+def get_client_id_path():
+    return "/iam/cis/{}/change_client_id".format(get_cis_environment())
+
+def get_client_secret_path():
+    return "/iam/cis/{}/change_service_client_secret".format(get_cis_environment())
+
+def get_url_dict():
+    cis_environment = get_cis_environment()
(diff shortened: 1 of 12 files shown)


e2e/helpers.py
```diff
@@ -0,0 +1,118 @@
...
+
+def get_client_id_path():
+    return "/iam/cis/{}/change_client_id".format(get_cis_environment())
+
+
+def get_client_secret_path():
+    return "/iam/cis/{}/change_service_client_secret".format(get_cis_environment())
+
+
+def get_url_dict():
```
gdestuynder: This should probably get the URls from the wk endpoint (this tests the wk endpoint as well, which is what people would probably use)

https://auth.allizom.org/.well-known/mozilla-iam in `api.endpoints` and `api.audience`

The well-known lib can get them also very easily

e2e/helpers.py
```diff
@@ -0,0 +1,118 @@
...
+    cis_environment = get_cis_environment()
+
+    if cis_environment == "development":
```
gdestuynder:
```suggestion
    # XXX Use well-known endpoint library for fetching URLs and audience
    if cis_environment == "development":
```
adding that here in case you don't actually use the wk lib, so that its clear one shouldnt copy these urls ;-)

PR #339, opened 2019-02-08
Add null_create profile.

fiji-flo:
- Add a null profile with publishers set according to create rules
- Add support to profile faker
- Fix primary_username encoding


gdestuynder:
@fiji-flo this needs a test :)

python-modules/cis_profile/tests/test_fake_profile.py
```diff
@@ -34,3 +37,13 @@ def test_with_and_without_uuid(self):
         assert a.uuid.value is not None
         b = fake_profile.FakeUser(seed=23, config=c_without_uuid)
         assert b.uuid.value is None
+
+    def test_null_create_profile(self):
+        empty_profile = profile.User()
+        create_profile = fake_profile.FakeUser(seed=1337, config=fake_profile.FakeProfileConfig().default().create())
+        update_profile = fake_profile.FakeUser(seed=1337, config=fake_profile.FakeProfileConfig().default())
+
+        with pytest.raises(exceptions.PublisherVerificationFailure):
```
fiji-flo: I was curious: shouldn't this return `False` instead of throwing an exception according to the doc string. I can change this in a follow-up PR.

andrewkrug: We actually made a conscious choice to have it raise a verification failure here.  It's handled in the upstream APIs where we do exactly what you describe and return a meaningful error.

gdestuynder: note the rationale so far has been

-  "if this is a big (eg security) issue if things go on here, raise an exception and stop the flow loudly"
- "if its optional, return a value such as true/false"

due to this publisher verification and signature verification failures raise exceptions

PR #350, opened 2019-02-27
fixes to ensure X-Ray tracing is working

andrewkrug:


andrewkrug:
If you're wondering why there's two commits... I forgot to setup signing on my new machine.

PR #360, opened 2019-03-06
initial commit of the notification formatter lib

andrewkrug:


python-modules/cis_notifications/setup.cfg
```diff
@@ -0,0 +1,19 @@
+[bumpversion]
+current_version = 0.3.8
```
gdestuynder: i think you meant 0.0.1

gdestuynder:
```suggestion
current_version = 0.0.1
```

PR #370, opened 2019-03-07
set content type to json

fiji-flo:


andrewkrug:
@fiji-flo best fix eva. :)

PR #381, opened 2019-03-13
default to staff and private for create

fiji-flo:


fiji-flo:
I'm not sure if we should pull those changes into the null profile (not create).

python-modules/cis_profile/cis_profile/data/user_profile_null_create.json
```diff
@@ -256,7 +256,7 @@
       "last_modified": "1970-01-01T00:00:00Z",
       "created": "1970-01-01T00:00:00Z",
       "verified" : true,
-      "display": null
+      "display": "staff"
```
gdestuynder: this should stay null for first/last name no?
or at least something non-staff - im guessing default is staff for staff and private for non-staff

gdestuynder:
yes if it goes into create it has to also go into null profile

python-modules/cis_profile/cis_profile/data/user_profile_null_create.json
fiji-flo: Yes. I think we settled on making this the skeleton for staff. I'm still in favor of dumping thins on the publishers :P

Do we wanna keep this for now or do you wanna put logic into the publishers? Or should I create different skeletons?

My main goal is to have phone book parity import for this week.

fiji-flo:
Okay, I set everything that should be set depending on staff/non-staff to null. Everything that's not (usually) set by hris or ldap defaults to private (or the most restrictive setting).

@gdestuynder could you take a look.

python-modules/cis_profile/cis_profile/data/user_profile_null.json
```diff
@@ -256,7 +256,7 @@
       "last_modified": "1970-01-01T00:00:00Z",
       "created": "1970-01-01T00:00:00Z",
       "verified" : true,
-      "display": null
+      "display": "staff"
```
gdestuynder:
```suggestion
      "display": null
```
this is last_name

python-modules/cis_profile/cis_profile/data/user_profile_null_create.json
```diff
@@ -1171,7 +1171,7 @@
         "last_modified": "1970-01-01T00:00:00Z",
         "created": "1970-01-01T00:00:00Z",
         "verified" : true,
-        "display": "ndaed"
```
gdestuynder: note I think these were actually OK as ndaed

gdestuynder:
last_name handling differs from first_name so i didnt approve

that said as i read it i wonder if we really need to do this? Since publishers would have to set their own defaults anyway.. 
I can add these to hris/ldap publishers (and auth0 when we write that)

gdestuynder:
will add to publishers instead as discussed

PR #399, opened 2019-03-27
fix missing hris attributes from certain users

gdestuynder:
note: this fixes hris.py by doing lowercase compares
publisher.py (the generic class) does not need this fix, it fetches user ids from CIS so it has the right id when doing filtering and comparisons

this also adds the filter for `display:null` because i forgot to branch out ./blame kang ; -)


gdestuynder:
https://stackoverflow.com/questions/1207406/how-to-remove-items-from-a-list-while-iterating if curious

PR #404, opened 2019-04-04
fix dev fixtures fix user_ids not being constructed correctly in some situations by always querying them

gdestuynder:


gdestuynder:
note: its on development

python-modules/cis_publisher/cis_publisher/publisher.py
```diff
@@ -34,13 +34,21 @@ def __init__(self, profiles, login_method, publisher_name, discovery_url=None):
...
         self.retry_delay = 1
+
+        # known_cis_users is the output of the Person API query
+        # known_cis_users_only_user_id is an array of user_id only
+        # known_cis_users_by_email is a dict that maps email: uid (instead of uid: email)
         self.cis_user_list = None
-        self.access_token = None
         self.known_cis_users = None
+        self.known_cis_users_by_email = {}
+        self.known_cis_users_only_user_id = []
```
andrewkrug: Just curious... should we standardize on dict or list here.

gdestuynder: yeah it could also be dict for both so that you can do the reverse lookup, that would be a bit nicer

gdestuynder: added

PR #408, opened 2019-04-08
Display null

andrewkrug:
Closes https://jira.mozilla.com/browse/EIS-729


python-modules/cis_profile_retrieval_service/cis_profile_retrieval_service/v2_api.py
```diff
@@ -138,6 +138,7 @@ def scope_to_display_level(scopes):
     if "display:all" in scopes:
         logger.debug("all display level in scope.")
         display_levels.append(DisplayLevel.NONE)
+        display_levels.append(DisplayLevel.NULL)
```
gdestuynder: aaah :)

python-modules/cis_profile_retrieval_service/cis_profile_retrieval_service/v2_api.py
```diff
@@ -155,6 +156,10 @@ def scope_to_display_level(scopes):
         logger.debug("authenticated display level in scope.")
         display_levels.append(DisplayLevel.AUTHENTICATED)
 
+    if "display:none" in scopes:
```
gdestuynder: does this mean auth0 should support `display:none` and `display:null` as scopes? if yes we need to add this in serverless/... (theres an auth0 "helper") and make sure the setup in auth0 support this

gdestuynder: note: right now it only supports `display:null`. we can also swap it for just `display:none` and tell people about it

PR #410, opened 2019-04-09
Use people.mozilla.org as DinoPark prod url

fiji-flo:


fiji-flo:
Let's keep this open until we settled. It's working right now.

PR #411, opened 2019-04-09
deactivation support

gdestuynder:
This supports HRIS user deactivation
its missing LDAP users deactivation for users who are not staff (who are contributors)
the later could be a separate work item actually

### DIFF
--- a/docs/PersonAPI.md
+++ b/docs/PersonAPI.md
@@ -132,6 +132,20 @@ Retrieve specific user profiles:
 Retrieve paginated list of all user profiles:
 - `/v2/users/`
 
+In this case you will receive a JSON document such as :
+```
+{
+  "Items": [profile_1, profile_2, ...],
+  "nextPage": None
+}
+```
+where `nextPage` will be set to a `user_id` if there are more pages to fetch.
+
+To fetch the next page simply call:
+- `/v2/users?nextPage=user_id_here`
+
+Until `nextPage` is set to `None`.
+
 Retrieve lists of users (not full profiles):
 - `/v2/users/id/all?[connectionMethod=ad]` (Returns all user ids for a specific login/connection method) By default this only includes active users.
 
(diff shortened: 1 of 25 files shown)


gdestuynder:
note: this currently has an issue i found during testing where it will deactivate certain accounts which arent staff and arent otherwise marked in some other way. discussing tomorrow.

gdestuynder:
this includes a revert for #419

docs/PersonAPI.md
```diff
@@ -129,7 +129,24 @@ Retrieve specific user profiles:
...
+- `/v2/users/`
+
+In this case you will receive a JSON document such as :
+```
+{
+  "Items": [profile_1, profile_2, ...],
+  "nextPage": None
+}
+```
+where `nextPage` will be set to a `user_id` if there are more pages to fetch.
```
fiji-flo: nextPage actually is an object:
```json
{ "id": "SOME_USER_ID" }
```
I fell for this and it drove me crazy to figure it out.

well-known-endpoint/tpl/mozilla-iam-publisher-rules
```diff
@@ -2,10 +2,10 @@
   "create": {
     "uuid": [ "cis" ],
     "user_id": [ "ldap", "hris", "access_provider" ],
-    "primary_username": [ "ldap", "hris", "cis", "access_provider" ],
+    "primary_username": [ "cis" ],
```
fiji-flo: :+1:

docs/PersonAPI.md
gdestuynder: ops didnt submit "review", i though i commented here
after some testing i seem to definitely be getting either none either a single user_id, not a structure

see the code also https://github.com/mozilla-iam/cis/pull/411/files#diff-3247471fbaab5e22352bc0bfc86e2063R244 this definitely gets all pages

gdestuynder:
something's odd with travis testing, local tests dont have this error / doesnt seem code related

gdestuynder:
non-squashed since we might need commit history here - in particular this reverts https://github.com/mozilla-iam/cis/pull/419 which we'll have to fix and pull back in

PR #413, opened 2019-04-10
import back-off code from cis_notifications for SSM fetches add in-memory caching

gdestuynder:


andrewkrug:
Suggest converting this to a proper boto3 waiter: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ssm.html#SSM.Client.get_waiter

PR #414, opened 2019-04-10
Filter active

andrewkrug:


python-modules/cis_identity_vault/cis_identity_vault/models/user.py
```diff
@@ -234,31 +269,47 @@ def all(self):
...
         logger.info('Getting segment: {} of total: {}'.format(segment, total_segments))
+
+        if connection_method:
+            expression_attr = {
+                ":id": {"S": connection_method}
+            }
+            filter_expression = "begins_with(id, :id)"
+
+        
+        if connection_method and active is not None:
```
gdestuynder: id generally `else` here (had to read this twice)

python-modules/cis_identity_vault/cis_identity_vault/models/user.py
```diff
@@ -331,3 +382,6 @@ def all_by_page(self, next_page=None, limit=25):
         else:
             response = self.table.scan(Limit=limit)
         return response
+
```
gdestuynder: lotsa lines!

PR #419, opened 2019-04-16
Flat profile instead of individual columns

andrewkrug:
This now represents the flattened profile as a singular map object.


andrewkrug:
This code makes this work btw:

```

PR #420, opened 2019-04-19
fix var init + Black

gdestuynder:


python-modules/cis_profile_retrieval_service/cis_profile_retrieval_service/v2_api.py
```diff
@@ -357,6 +357,8 @@ def get(self):
             # This must be a pre filtering check because mutation is real.
             if v2_profile.active.value == active:
                 allowed_in_list = True
+            else:
+                allowed_in_list = False
```
gdestuynder: this var was otherwise not initialized in all cases so could end up triggering a traceback when used 
this would also not reset the variable in the loop which is a real bug as well

april: Wouldn't it be easier to write it as:

`allowed_in_list = v2_profile.active.value == active`

gdestuynder: its definitelly prettier but i dont know how readable it is

april: Either this or `allowed_in_list = True if v2_profile.active.value == active else False`, which is probably more readable.

april: Also looking at this a bit more, what happens if neither of the two if statements work? Like if args.active("active") is None but the value is false? The code would crash here.

gdestuynder: yeah its weird logic - not part of the PR but might as well fix it. i dont think it'd crash because it accounts for all actual cases, but its hard to read and makes it easy to add code that will break this.

gdestuynder: (which is now changed)

gdestuynder:
^ this needs more mocks :)

PR #422, opened 2019-04-23
fix backoff fix merge difference return value improve cache settings

gdestuynder:
fix https://jira.mozilla.com/browse/DP-1262

### DIFF
--- a/python-modules/cis_crypto/cis_crypto/secret.py
+++ b/python-modules/cis_crypto/cis_crypto/secret.py
@@ -106,7 +106,7 @@ class AWSParameterstoreProvider(object):
             logger.debug("Returning memory-cached version of uuid_salt")
             return self._cache["uuid_salt"]
 
-        while result is None or retries != 0:
+        while result is None:
             try:
                 ssm_path = self.config("secret_manager_ssm_uuid_salt", namespace="cis", default="/iam")
                 ssm_response = self.ssm_client.get_parameter(Name=ssm_path, WithDecryption=True)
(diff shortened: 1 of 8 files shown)


serverless-functions/hris_publisher/serverless.yml
```diff
@@ -103,7 +108,7 @@ functions:
     handler: handler.handle
     events:
       - schedule:
-          rate: rate(2 hours)
+          rate: rate(6 hours)
```
gdestuynder: this was changed as per HR's demand

python-modules/cis_crypto/tests/test_secret.py
```diff
@@ -51,21 +51,23 @@ def test_ssm_provider(self):
...
+    #        client = boto3.client("ssm", region_name="us-west-2")
+    #        client.put_parameter(
+    #            Name="/salty",
+    #            Description="A fake salt",
+    #            Value="fancyseruresalt",
+    #            Type="SecureString",
+    #            KeyId="alias/aws/ssm",
+    #        )
+    #        salt = secret.AWSParameterstoreProvider().uuid_salt()
+    #        assert salt is not None
```
gdestuynder: i could mock uuid salt like i do elsewhere but it makes for a test of no value in this case, so its commented out until mock_ssm does the right thing

python-modules/cis_crypto/cis_crypto/secret.py
```diff
@@ -100,7 +106,7 @@ def uuid_salt(self):
             logger.debug("Returning memory-cached version of uuid_salt")
             return self._cache["uuid_salt"]
 
-        while result is None or retries == 0:
+        while result is None or retries != 0:
```
gdestuynder:
```suggestion
        while result is None:
```

python-modules/cis_crypto/tests/test_secret.py
gdestuynder: @andrewkrug

PR #424, opened 2019-04-30
Fix flat profile

andrewkrug:


python-modules/cis_profile/cis_profile/profile.py
```diff
@@ -294,9 +294,10 @@ def flatten(attrs, field=None):
                 # Skip "schema"
                 if isinstance(attrs[f], str):
                     continue
-                if not set(["value", "values"]).isdisjoint(set(attrs[f])):
+                elif not set(["value", "values"]).isdisjoint(set(attrs[f])):
                     res = attrs[f].get("value", attrs[f].get("values"))
-                    if res is not None and res != "":
+
+                    if res is not None and res != "" and isinstance(res, str):
```
gdestuynder: doesnt that skip all list and dict items?
ie. that would never add "access_information.ldap.values = {...}"

gdestuynder: actually i think that yes it does:

```
self = <tests.test_fake_profile.TestFakeProfile object at 0x7f7f04186be0>
    def test_fake_dynamo_flat_dict_output(self):
        # This profile must have ldap groups and staff data
        p = fake_profile.FakeUser(seed=44)
        ddb = p.as_dynamo_flat_dict()
        deserializer = TypeDeserializer()
        res = {k: deserializer.deserialize(v) for k, v in ddb.items()}
>       assert len(res["access_information"]["ldap"]) > 0
E       KeyError: 'ldap'
tests/test_fake_profile.py:59: KeyError
```

PR #425, opened 2019-04-30
Only return active users in /v2/user/*

fiji-flo:
- add `active` as query paramter to retrieve inactive users


fiji-flo:
I had some issues with `make test-tox` :crossed_fingers:

gdestuynder:
lookin' good

PR #431, opened 2019-06-04
move key ownership to ldap

fiji-flo:


andrewkrug:
This only sort of works for now.  If LDAP owns the keys how do we guarantee community can add keys later on?  Or do these eventually become owned wholly by DinoPark?

fiji-flo:
Yes. Plan is have this for All-Hands and then move ownership to DinoPark.

gdestuynder:
@fiji-flo i gotta deploy manually this right?
let me know when

PR #434, opened 2019-06-07
Fix e2e tests.

fiji-flo:


e2e/helpers.py
```diff
@@ -121,11 +121,7 @@ def ensure_appropriate_publishers_and_sign(self, fake_profile, condition):
             else:
                 if attr != "schema" and attr in complex_structures:
                     for k in temp_profile[attr]:
-                        if attr == "access_information":
-                            successful_random_publisher = random.choice(publisher_rules[condition][attr][k])
-
-                        if attr == "staff_information" or attr == "identities":
-                            successful_random_publisher = random.choice(publisher_rules[condition][attr])
+                        successful_random_publisher = random.choice(publisher_rules[condition][attr][k])
```
gdestuynder: @fiji-flo you're sure this does not break anything even when you run the e2e several times?

gdestuynder: @fiji-flo

andrewkrug: @fiji-flo can you comment on this or close?

PR #436, opened 2019-07-25
initial support for postgres in parity with dynamodb

andrewkrug:
This lands the initial set of foundations to replicate data from DynamoDb to postgresql.  I've maintained the top level keys and serialized the rest of the user profile to a JSON column.  Note: Only the methods that existed prior to now for search have been implemented.  The next iteration will include several more generic interfaces for querying groups and HR data.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,3 +51,8 @@ build-ci-container:
 .PHONY: login-to-ecr upload-ci-container
 push-ci-container:
 	docker push 320464205386.dkr.ecr.us-west-2.amazonaws.com/custom-codebuild-cis-ci
+
+.PHONY: test
+test:
+	docker-compose build
+	docker-compose run tester bash -c '/root/utils/fake-creds.sh && source /root/.bashrc && make -j4 test-tox'
--- a/python-modules/cis_identity_vault/cis_identity_vault/models/user.py
+++ b/python-modules/cis_identity_vault/cis_identity_vault/models/user.py
@@ -216,7 +216,8 @@ class Profile(object):
             if user_profile["sequence_number"] is None:
                 user_profile["sequence_number"] = str(uuid.uuid4().int)
 
-            cis_profile_user_object = User(user_structure_json=json.loads(profile["profile"]))
+            # XXX TBD cover this with tests.  Currently dynalite does not support tests for transactions.
+            cis_profile_user_object = User(user_structure_json=json.loads(user_profile["profile"]))
             transact_item = {
                 "Put": {
                     "Item": {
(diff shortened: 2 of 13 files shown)


andrewkrug:
FYI... this got a little muddy when I rebased master into the branch in order to squash my commits for travis tests.

Makefile
```diff
@@ -51,3 +51,8 @@ build-ci-container:
 .PHONY: login-to-ecr upload-ci-container
 push-ci-container:
 	docker push 320464205386.dkr.ecr.us-west-2.amazonaws.com/custom-codebuild-cis-ci
+
+.PHONY: test
+test:
+	docker-compose build
+	docker-compose run tester bash -c '/root/utils/fake-creds.sh && source /root/.bashrc && make -j4 test-tox'
```
gdestuynder: because this can make a huge difference in this case - quick optimization:
```
make -j $(getconf _NPROCESSORS_ONLN)
```

or maybe nicer:
```
NPROC = $(shell getconf _NPROCESSORS_ONLN)

....
... make -j $(NPROC)
```

python-modules/cis_identity_vault/cis_identity_vault/models/user.py
```diff
@@ -417,3 +430,98 @@ def all_by_page(self, next_page=None, limit=25):
...
+    def find_by_username(self, primary_username):
+        try:
+            user = (
+                self.session.query(rds.People)
+                .filter(rds.People.profile[("primary_username", "value")].astext == primary_username)
+                .one()
+            )
+        except NoResultFound:
+            user = None
+        return user
```
gdestuynder: man this is so much easier to read than DDB ;P

python-modules/cis_profile/cis_profile/profile.py
```diff
@@ -434,7 +434,7 @@ def verify_can_publish(self, attr, attr_name, parent_name=None, previous_attribu
             try:
                 allowed_updators = rules["update"][parent_name][attr_name]
             except TypeError:  # This is not access_information, this is identities or staff_information
-                allowed_updators = rules["create"][parent_name]
+                allowed_updators = rules["update"][parent_name]
```
gdestuynder: @andrewkrug do you remember what this fixes? the fix looks logical but, i'm wondering how that was found / just in case this impact other code

gdestuynder: or maybe that was already there and merge shenanigans

andrewkrug: This wasn't part of my work for this.  This came with the rebase.

PR #437, opened 2019-08-01
GroupAPI spec 0

gdestuynder:
This is mainly putting the docs and current reference work in one place


andrew-sunada:
Is there already implementation or is this just the api setup?

gdestuynder:
There is no implementation, the RFCs are used for implementation.
In this case its likely that we load the OpenAPI spec so it would match implementation 1 to 1 - but right now this is just an RFC (as in Request For Comment)

gdestuynder:
since its just doc / no code and we need it in for iteration im going to merge this

PR #439, opened 2019-08-23
add mozilliansorg-groups-publisher

fiji-flo:


python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
```diff
@@ -0,0 +1,102 @@
...
+        return None
+
+
+class MozilliansorgGroupsPublisher:
+    def __init__(self):
+        self.secret_manager = cis_publisher.secret.Manager()
+        self.publisher = cis_publisher.Publish([], login_method=None, publisher_name="mozilliansorg")
+
+    def publish(self, event):
+        updates = [update for update in map(MozilliansorgGroupUpdate.from_record, event.get("Records", [])) if update]
```
gdestuynder: add a comment on what this does (both lines 67 and 68)

python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
```diff
@@ -0,0 +1,102 @@
...
+        return None
+
+
+class MozilliansorgGroupUpdate:
+    def __init__(self, typ, user_id, groups=[]):
+        self.typ = typ
+        self.user_id = user_id
+        self.groups = groups
+
+    def from_record(record):
```
gdestuynder: docstring this and other methods

python-modules/cis_publisher/cis_publisher/publisher.py
```diff
@@ -151,16 +151,19 @@ def post_all(self, user_ids=None):
...
-        retries = 0
-        access_token = self._get_authzero_token()
-
         # Existing users (i.e. users to update) have to be passed as argument
         if user_id in self.known_cis_users_by_user_id:
             qs = "/v2/user?user_id={}".format(user_id)
         # New users do not
         else:
             qs = "/v2/user"
+        self._really_post_with_qs(user_id, qs, profile, failed_users)
```
gdestuynder: can't this just replace `_really_post` instead?

fiji-flo: I'd rather do that as follow up to make it easily revertible. We're not using `qs` in `_really_post, but we're passing it in several cases. What `_really_post_with_qs` does not do is the implicit checking for known users. I'd propose I refactor this in follow up and have a `_really_post_with_known_userid` and a `_really_post` method or something like that.

python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
```diff
@@ -0,0 +1,122 @@
...
+        update_profile.active = current_profile.active
+        update_profile.access_information.mozilliansorg["values"] = {group: None for group in update.groups}
+        if (
+            update_profile.access_information.mozilliansorg["values"]
+            == current_profile.access_information.mozilliansorg["values"]
+        ):
+            logger.info("No change in mozilliangsorg access information. Skipping.")
+            return None
+        if not update_profile.access_information.mozilliansorg.metadata.display:
+            update_profile.access_information.mozilliansorg.metadata.display = "ndaed"
```
fiji-flo: @gdestuynder changed this to _ndaed_ according to Hermina.

python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
```diff
@@ -0,0 +1,122 @@
...
+            logger.info("No profile for {}, skipping!".format(update.user_id))
+            return None
+
+        logger.info("received profile for {}".format(current_profile.user_id.value))
+        update_profile = cis_profile.profile.User()
+        update_profile.user_id = current_profile.user_id
+        update_profile.active = current_profile.active
+        update_profile.access_information.mozilliansorg["values"] = {group: None for group in update.groups}
+        if (
+            update_profile.access_information.mozilliansorg["values"]
```
fiji-flo: @gdestuynder checking id anything changed now.

python-modules/cis_publisher/cis_publisher/publisher.py
gdestuynder: yeah sounds better

python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
```diff
@@ -0,0 +1,122 @@
...
+class MozilliansorgGroupUpdate:
+    def __init__(self, typ, user_id, groups=[]):
+        self.typ = typ
+        self.user_id = user_id
+        self.groups = groups
+
+    def from_record(record):
+        """
+        Constructs a MozilliansGroupUpdate from a raw sqs record.
+        Return a MozilliansGroupUpdate or None to error gracefully.
```
gdestuynder: Something like this would be good additionally, as per other pieces of code we have (you can look at https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings for inspiration though its not super well defined AFAIK)

```
Arguments:
typ -- it's ... (i dont actually know yet, ive to find the function call and figure out what that is)
user_id -- str the IAM user id
groups -- list of mozilliansorg groups for that user_id
```

This helps understanding what the code does a whole lot without manually tracing back the function calls

note: its a good habit for most functions, though the ones that are really straight forward dont really matter too much (like `unpack_list_from_string` or other small helpers)

same for classes

python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
```diff
@@ -0,0 +1,122 @@
...
+            body = json.loads(body)
+        except json.JSONDecodeError as e:
+            logger.error("Invalid JSON in body: {}".format(e))
+            return None
+        user_id = get_nested(body, "dynamodb", "Keys", "user_id", "S")
+        new_image = get_nested(body, "dynamodb", "NewImage")
+        typ = get_nested(body, "eventName")
+        payload_user_id = get_nested(new_image, "user_id", "S")
+        if user_id != payload_user_id:
+            logger.error("missmatching user_ids {} {}".format(user_id, payload_user_id))
```
gdestuynder:
```suggestion
            logger.error("mismatching user_ids {} {}".format(user_id, payload_user_id))
```

nitpick - typo

python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
gdestuynder: im not sure what your comment means - but this looks alright. id also skip sending the same way.
you might want to call `sorted` for safety though as the array order is not guaranteed, i.e.

`if sorted(update_profile.access_information.mozilliansorg["values"]) == sorted(...)`

serverless-functions/mozilliansorg_groups_publisher/handler.py
```diff
@@ -0,0 +1,30 @@
...
+patch_all()
+mozilliansorg_groups = cis_publisher.mozilliansorg_groups.MozilliansorgGroupsPublisher()
+
+
+def setup_logging():
+    logger = logging.getLogger()
+    for h in logger.handlers:
+        logger.removeHandler(h)
+    h = logging.StreamHandler(sys.stdout)
+    h.setFormatter(cis_logger.JsonFormatter(extra={"hostname": socket.gethostname()}))
```
gdestuynder: I forgot if we do this elsewhere, but the extra should probably be defaulted in `cis_logger`

fiji-flo: extra defaults to `{}` if I see this right.

PR #444, opened 2019-09-30
WIP on Postgres Support

andrewkrug:


ci/Dockerfile
```diff
@@ -23,5 +27,18 @@ RUN npm install dynalite
 RUN mkdir /root/utils
 COPY fake-creds.sh /root/utils/
 RUN chmod 700 /root/utils/fake-creds.sh
+RUN wget https://ftp.postgresql.org/pub/source/v11.1/postgresql-11.1.tar.gz
```
gdestuynder: was checking but it looks like they dont sign their releases :(

python-modules/cis_aws/package-lock.json
```diff
@@ -0,0 +1,810 @@
+{
```
gdestuynder: I think we should probably not store this file, so that its recreated with up to date deps each time.
for ex the current one had an dependency loop with leveldown, but that would also ensure vulns to be patched automatically

python-modules/cis_identity_vault/cis_identity_vault/models/rds.py
```diff
@@ -1,7 +1,16 @@
...
+logger = logging.getLogger(__name__)
+
+
+try:
+    from sqlalchemy import Column, Integer, Text
+    from sqlalchemy.dialects.postgresql import JSON, JSONB
+    from sqlalchemy.orm import sessionmaker
+    from sqlalchemy.ext.declarative import declarative_base
+except ImportError:
+    logger.error('Postgresql support not available.  Try installing psycopg2.')
```
gdestuynder:
```suggestion
    logger.error('Postgresql support not available. Try installing psycopg2. Error: {}'.format(e))
```

python-modules/cis_identity_vault/cis_identity_vault/models/rds.py
```diff
@@ -1,7 +1,16 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+try:
+    from sqlalchemy import Column, Integer, Text
+    from sqlalchemy.dialects.postgresql import JSON, JSONB
+    from sqlalchemy.orm import sessionmaker
+    from sqlalchemy.ext.declarative import declarative_base
+except ImportError:
```
gdestuynder:
```suggestion
except ImportError as e:
```

python-modules/cis_identity_vault/cis_identity_vault/models/rds.py
gdestuynder: just good practice in case it ever goes wrong

python-modules/cis_identity_vault/cis_identity_vault/vault.py
```diff
@@ -324,3 +328,11 @@ def table(self):
         metadata = rds.Base.metadata
         metadata.bind = self.engine()
         return metadata.tables.get("people")
+    
+    def find_or_create(self):
+        try:
+            self.table()
```
gdestuynder:
```suggestion
            return self.table()
```

right?

python-modules/cis_identity_vault/setup.py
```diff
@@ -6,7 +6,7 @@
 with open("README.md", "r") as fh:
     long_description = fh.read()
 
-requirements = ["everett", "everett[ini]", "configobj", "boto", "boto3", "botocore", "sqlalchemy", "psycopg2"]
+requirements = ["everett", "everett[ini]", "configobj", "boto", "boto3", "botocore", "sqlalchemy", "sqlalchemy_utils"]
```
gdestuynder: as per error msg during import, it sounds like this should include both sqlalchemy_utils and psycopg2

python-modules/cis_identity_vault/tox.ini
```diff
@@ -15,4 +15,5 @@ deps=
   tox-run-before
   ../cis_profile
   ../cis_crypto
+  /psycopg2-2.8.3
```
gdestuynder: any issue with getting it from https://pypi.org/project/psycopg2/ instead? (same for the docker image build) and get the "latest version" every time?

python-modules/cis_postgresql/cis_postgresql/exchange.py
```diff
@@ -0,0 +1,138 @@
...
+    Stores the profiles in the relational store as valid structured JSON.
+    Supports batch only interaction.
+    """
+
+    def to_postgres(self, profiles):
+        rds_vault = user.ProfileRDS()
+        results = []
+        for _ in profiles:
+            results.append(rds_vault.find_or_create(_))
+        logger.info(f'{len(results)} profiles have been written to the postgresql identity vault.')
```
gdestuynder: nit: mixed syntax style :(

python-modules/cis_postgresql/cis_postgresql/execute.py
```diff
@@ -0,0 +1,158 @@
...
+from sqlalchemy.dialects.postgresql import JSON, TEXT
+
+def attr_single_or_multi(attr):
+    """[return value or values for an attribute for use in queries]
+    
+    Arguments:
+        attr {string} -- [name of the attribute]
+    """
+
+    values = [
```
gdestuynder: other pieces of code just try if its values vs value. either way it works, just wanted to mention it

python-modules/cis_postgresql/setup.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from setuptools import setup, find_packages
+
+with open("README.md", "r") as fh:
```
gdestuynder: this no longer works, pip warehouse only support retructured text/rst (not markdown)
one quick fix is to just... not have a description or just a link to our git repos. a better fix would be to convert the text.

serverless-functions/Makefile
```diff
@@ -22,39 +23,15 @@ all:
 
 .PHONY:clean-layer
 clean-layer:
-	rm -rf $(ROOT_DIR)/layer/*
-	rm -rf $(ROOT_DIR)/build/*
-	rm -f $(ROOT_DIR)/cis.zip
-
-.PHONY:layer
-layer:
```
gdestuynder: was this target (`layer`) not used?

python-modules/cis_postgresql/setup.py
andrewkrug: I'll open an issue to fix separately.

serverless-functions/Makefile
andrewkrug: Correct it was not used.

PR #446, opened 2019-09-30
enable task add user which can trigger the lambda with a long term Access key, for hooks

gdestuynder:


serverless-functions/auth0_publisher/serverless.yml
```diff
@@ -115,9 +115,49 @@ functions:
...
+                    - "lambda:InvokeAsync"
+                  Resource:
+                    - arn:aws:lambda:us-west-2:320464205386:function:auth0-publisher-${self:custom.auth0PublisherStage}-handler
+    AZRunnerAccessKey:
+      Type: AWS::IAM::AccessKey
+      Properties:
+        Serial: 20190930
+        Status: Active
+        UserName: !Ref AZRunnerUser
+    BasicParameter:
```
gdestuynder: FYI @gene1wood - @andrewkrug suggested using SSM instead of outputs, which seems sane. Sadly, this does not work with SecureString, but its better than nothing (in case you want to change that in your AWS CLI Federation code)

PR #448, opened 2019-10-02
remove reserved concurrency

fiji-flo:
for moziliansorg groups publisher


serverless-functions/mozilliansorg_groups_publisher/serverless.yml
```diff
@@ -79,4 +79,3 @@ functions:
     timeout: 30
```
andrewkrug: @fiji-flo update timeout to 900

PR #454, opened 2019-10-16
staff_information.staff display value to be public

gdestuynder:
change publishing default for staff_information.staff display value to be public instead of staff See https://jira.mozilla.com/browse/IAM-292


gdestuynder:
this allows the "S" to appear on pictures or identicons

PR #459, opened 2019-10-22
Concurrency2

andrewkrug:


Makefile
```diff
@@ -33,7 +33,8 @@ deploy-shell: login-to-ecr
 
 .PHONY: build
 build:
-	docker-compose run tester make -C serverless-functions layer-codebuild STAGE=$(STAGE)
+	docker-compose run tester make -C serverless-functions build-without-publish STAGE=$(STAGE)
+	make -C serverless-functions publish-layer STAGE=$(STAGE)
```
gdestuynder: Maybe we should separate building and publishing. up to you, its just a suggestion
```suggestion
.PHONY: publish
publish:
	make -C serverless-functions publish-layer STAGE=$(STAGE)
```

python-modules/cis_identity_vault/cis_identity_vault/models/user.py
```diff
@@ -350,6 +351,17 @@ def _get_segment(self, segment=0, total_segments=10, connection_method=None, act
...
             )
             users.extend(response["Items"])
+
+        return users
+
+    async def _resultifier(self, pool_size, connection_method, active):
+        users = []
+        segment_result = None
+        for x in range(0, pool_size):
+            segment_result = await self._get_segment(
```
gdestuynder: is this really faster, if we await on the longest operation?

serverless-functions/profile_retrieval/package.json
```diff
@@ -11,6 +11,7 @@
   "license": "ISC",
```
gdestuynder: i wonder if we should have package.json just recreated by sls every time to avoid pinning deps

andrewkrug: The only reason there's package.json in there at all is for serverless-framework plugins.

python-modules/cis_identity_vault/cis_identity_vault/models/user.py
andrewkrug: Yes.

gdestuynder: nope! ;-) https://github.com/mozilla-iam/cis/pull/463

PR #460, opened 2019-10-22
fix token cache

fiji-flo:


python-modules/cis_notifications/cis_notifications/event.py
```diff
@@ -72,21 +72,21 @@ def send(self, notification):
...
-                self.secret_manager.secretmgr_store("az_access_token", self.access_token_dict)
-            else:
-                logger.info("Re-using cached access token")
+            self.access_token_dict["exp"] = int(self.access_token_dict.get("exp", -99999999999))
+        # 10s leeway
+        if not self.access_token or int(self.access_token_dict["exp"]) < time.time() - 10:
+            logger.info("Access token has expired, refreshing")
+            authzero = self._get_authzero_client()
+            self.access_token_dict = authzero.exchange_for_access_token()
+            self.access_token_dict["exp"] = time.time() + int(self.access_token_dict("expires_in"), 60)
```
gdestuynder: seems wrong still - this seems more like it:
```suggestion
            self.access_token_dict["exp"] = time.time() + int(self.access_token_dict.get("expires_in", 60))
```

python-modules/cis_notifications/cis_notifications/event.py
```diff
@@ -72,21 +72,21 @@ def send(self, notification):
...
-                logger.info("Access token has expired, refreshing")
-                authzero = self._get_authzero_client()
-                self.access_token_dict = authzero.exchange_for_access_token()
-                self.access_token_dict["exp"] = time.time() + int(self.access_token_dict("expires_in"), 60)
-                self.secret_manager.secretmgr_store("az_access_token", self.access_token_dict)
-            else:
-                logger.info("Re-using cached access token")
+            self.access_token_dict["exp"] = int(self.access_token_dict.get("exp", -99999999999))
+        # 10s leeway
+        if not self.access_token or int(self.access_token_dict["exp"]) < time.time() - 10:
```
gdestuynder:
```suggestion
        elif not self.access_token or int(self.access_token_dict["exp"]) < time.time() - 10:
```

fiji-flo: `elif` won't work but my thing is broken, too.

PR #464, opened 2019-10-22
fix caching of access_token (logic errors) + performance

gdestuynder:
replaces https://github.com/mozilla-iam/cis/pull/460

### DIFF
--- a/python-modules/cis_notifications/cis_notifications/event.py
+++ b/python-modules/cis_notifications/cis_notifications/event.py
@@ -79,13 +79,13 @@ class Event(object):
 
             # Check if what we had in secrets is still valid!
             # This includes 10s leeway for clock sync issues
-            if int(self.access_token_dict["exp"]) < time.time() - 10:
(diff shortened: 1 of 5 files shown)


gdestuynder:
Added a commit just for profile retrieval perf

python-modules/cis_notifications/cis_notifications/event.py
```diff
@@ -72,17 +72,20 @@ def send(self, notification):
 
         [return] Dictionary of status codes by publisher.
         """
+        # Not in-memory access token?
```
fiji-flo: This can still send an invalid token while performing sending a large batch? We batch up to 100 messages which can take more than the 10s grace.

gdestuynder: the function can take up to <max time set in serverless.yml up to 15min> so potentially we could do a "up to 15min+10" of leeway

gdestuynder: added

PR #473, opened 2019-11-12
change profile defaults

gdestuynder:


python-modules/cis_publisher/cis_publisher/publisher.py
```diff
@@ -487,12 +487,21 @@ def filter_known_cis_users(self, profiles=None, save=True):
...
-                        p.__dict__[pfield] = null_user.__dict__[pfield]  # reset
-                # XXX filter sub-fields too on create? ["identities", "staff_information", "access_information"]
+                    # XXX filter more sub-fields on create? ["staff_information", "access_information"]
+                    # Refactor me to be recursive (just like above code)
+                    if pfield == "identities":
+                        for subpfield in p.__dict__[pfield]:
+                            f = p.__dict__[pfield][subpfield]
+                            if "value" in f.keys() and f["value"] is None:
+                                p.__dict__[pfield][subpfield] = null_user.__dict__[pfield][subpfield]  # reset
+                            elif "values" in f.keys() and (f["values"] is None or len(f["values"]) == 0):
```
fiji-flo: That didn't change, but we usually don't consider an empty list to be nothing.

gdestuynder: I do during publishing because it's the original source of data - i.e. empty list is allowed after this
Basically these publishers have different rules than Dinopark, I think it makes sense. But if Dinopark used a publisher like this, it would need its own code (but also, it won't use this :)

PR #483, opened 2019-12-03
Fix case where certain users are not imported and upgrade to p3.8 (required for fix to land)

gdestuynder:
tested in testing

- fix case where the user has both Mozilla-LDAP and another connection fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1595749
- upgrade to python3.8 as some libraries in 3.7 are no longer forward compatible and make the deploy and tests fail. AWS released 3.8 today..

This is what a run with no-cache now looks like (it attempts the user in the blacklisted list of connections, and ends up with nothing to post, so bails - this is also more efficient obviously, as it previously tried to post a profile that it knew would fail to be accepted)

```
[2019-12-03T16:12:13-08:00] (2019/12/04/[$LATEST]6ef71aa6598943deb04174f2565ec41a) { "asctime": "2019-12-04T00:12:13Z+0000", "filename": "publisher.py", "funcName": "post_all", "hostname": "169.254.84.189", "levelname": "INFO", "lineno": 101, "message": "Requesting a specific list of user_id's to post ['ad|Mozilla-LDAP|TEST'] (total user_ids: 1, total profiles: 0)", "module": "publisher", "name": "cis_publisher.publisher", "processName": "MainProcess", "threadName": "MainThread", "timestamp": "2019-12-04T00:12:13Z+0000" }
[2019-12-03T16:12:13-08:00] (2019/12/04/[$LATEST]6ef71aa6598943deb04174f2565ec41a) { "asctime": "2019-12-04T00:12:13Z+0000", "filename": "publisher.py", "funcName": "post_all", "hostname": "169.254.84.189", "levelname": "INFO", "lineno": 121, "message": "After filtering, we have 0 user profiles to post", "module": "publisher", "name": "cis_publisher.publisher", "processName": "MainProcess", "threadName": "MainThread", "timestamp": "2019-12-04T00:12:13Z+0000" }
```

Note, there is a bug in a dependency lib in p3.8:
```
[2019-12-03T16:17:07-08:00] (2019/12/04/[$LATEST]6b9881b47b9a4b3d9d93b4e59b2667d1) /opt/python/lib/python3.8/site-packages/jmespath/visitor.py:34: SyntaxWarning: "is" with a literal. Did you mean "=="?
[2019-12-03T16:17:07-08:00] (2019/12/04/[$LATEST]6b9881b47b9a4b3d9d93b4e59b2667d1)   elif y is 0 or y is 1:
```

Since it's just a warning, i let it go. This should be fixed in jmespath or it's caller (see https://github.com/jmespath/jmespath.py/issues/187)


gdestuynder:
Add some p3.8 fixes to allow tests to pass
Added some logging that was previously removed (functions that issue zero logs aren't great)
Allow publisher to activate all users (missing part of the fix)

python-modules/cis_publisher/cis_publisher/auth0.py
```diff
@@ -233,12 +233,10 @@ def fetch_az_users(self, user_ids=None):
...
             t = " OR "
         az_query = az_query + ")"
-        az_query = az_query + " AND NOT ("
-        t = ""
-        for azc in self.az_blacklisted_connections:
-            az_query = az_query + t + 'identities.connection:"{}"'.format(azc)
-            t = " OR "
-        az_query = az_query + ")"
+        # NOTE XXX: There is no way to tell auth0's ES "don't include matches where the first identity.connection is a
+        # blacklisted connection", so we do this instead. This 100% relies on auth0 user_ids NOT being opaque,
```
arcrose: Is there any documentation that can be linked to that guarantees that these IDs _aren't_ opaque for now?  It would be useful to reference here (i.e. link to directly in this comment) in case a bug appears here in the future.

gdestuynder: No :( From a data point of view they should be considered opaque
From a pragmatic reality they won't be with auth0. I would hope that if they change it we have another way to identify these accounts - otherwise, the code would simply "do more work than it needs", i.e., it will match wrongly, convert the auth0 users into our standardized user profiles, then a function checks again if it's importing the wrong users and drop them (that function is already there as a safety)

PR #484, opened 2019-12-17
fix date time format

fiji-flo:


python-modules/cis_publisher/cis_publisher/hris.py
```diff
@@ -407,7 +407,7 @@ def strbool_convert(v):
                 continue
 
             p = cis_profile.User()
-            ts_now = time.strftime("%Y-%m-%dT:%H:%M:%SZ")
+            ts_now = time.strftime("%Y-%m-%dT%H:%M:%S.000Z", time.gmtime())
```
fiji-flo: there was a `:` after the `T` :fire:

PR #489, opened 2020-01-17
Fix deactivation

gdestuynder:
API/Dynamo now sends back profiles as: {"id":... "profile": {"user_id":...}
instead of {"user_id":...}

Use this syntax and make it a little bit more reliable and easier to
debug if any format change occurs in the future. Turn on DEBUG logging
to see the additional messages

### DIFF
--- a/python-modules/cis_publisher/cis_publisher/hris.py
+++ b/python-modules/cis_publisher/cis_publisher/hris.py
@@ -134,8 +134,14 @@ class HRISPublisher:
                 # Convert as needed to a dict
                 try:
                     profile = profile.as_dict()
-                except (TypeError, AttributeError):
-                    pass
+                except (TypeError, AttributeError) as e:
+                    logger.debug(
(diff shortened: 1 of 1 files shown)


gdestuynder:
note this ran through tests and was merged as a hotfix in the code base

python-modules/cis_publisher/cis_publisher/hris.py
```diff
@@ -134,8 +134,14 @@ def deactivate_users(self, publisher, profiles, report):
                 # Convert as needed to a dict
                 try:
                     profile = profile.as_dict()
-                except (TypeError, AttributeError):
-                    pass
+                except (TypeError, AttributeError) as e:
```
fiji-flo: any clue why this can fail?

gdestuynder: changes in API replies from CIS to the publisher

PR #493, opened 2020-02-12
fix inactive users

fiji-flo:


python-modules/cis_publisher/cis_publisher/publisher.py
```diff
@@ -354,6 +354,7 @@ def get_known_cis_userids_paginated(self, include_inactive=False):
             nextPage = response_json.get("nextPage")
 
         if include_inactive:
+            nextPage = ""
```
fiji-flo: After exiting the previous loop `nextPage` was set to `None`. Due to this the loop in this branch is never executed.

april: This looks fine, but I really am not a huge fan of this nomenclature of setting `nextPage = ""`. `""` shouldn't be a substitute for truthiness or an integer.

Honestly, I would probably structure this entire mess as `while True` and then have

```
if response_json.get("nextPage") is None:
    break
````

Which is far more legible and probably would have kept this bug from occurring in the first place.

Anyways, that's a very longwinded way of saying that this is fine.

fiji-flo: I do agree here. But I wanted to keep the fix minimal for now. We should clean this up though.

Issue #492, opened 2020-02-04
Tests failing because botocore can't connect to endpoint


gene1wood:
`E           botocore.exceptions.EndpointConnectionError: Could not connect to the endpoint URL: "http://localhost:32078/"`

Tests began failing around the time #489 was merged.

I'm not sure what the cause is but tests now fail.

floatingatoll:
Tests no longer fail in this manner. I de-paralleled them and found a missing dependency in one of the modules that, when added, repaired the "start a network listener" function in the tests. Fixed by #529.

PR #496, opened 2020-04-14
Initial doc on how configuration works

gene1wood:


gene1wood:
Tests failed because of #492

PR #499, opened 2020-04-21
merge pmo groups with mozillians groups

fiji-flo:


april:
I’ll take a look at this tomorrow. Wasn’t feeling super great today.

python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
```diff
@@ -126,3 +125,25 @@ def _prepare_update(self, update):
             logger.critical("Profile data signing failed for user {}: {}".format(update.user_id, e))
             return None
         return update_profile
+
+    @staticmethod
+    def _update_groups(current_groups, mozillians_groups_update):
+        """
+        Update current gropus based on a mozillians groups update. PMO groups (value == "") overwrite mozillians groups.
```
april, line 132: "gropus"

python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
```diff
@@ -126,3 +125,25 @@ def _prepare_update(self, update):
...
+
+        @current_groups: current groups dict from profile v2
+        @mozillians_gropus_update: list of mozillians groups
+        @return: new groups dict or None if nothing changed
+        """
+        if mozillians_groups_update is None:
+            return None
+        updated = {group: None for group in mozillians_groups_update}
```
april, line 140: Can you help me understand why it's `group: None` like this?

python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
```diff
@@ -126,3 +125,25 @@ def _prepare_update(self, update):
...
+        if not current_groups:
+            return updated
+        pmo_groups = {name: meta for name, meta in current_groups.items() if meta is not None}
+
+        updated.update(pmo_groups)
+
+        if updated == current_groups:
+            return None
```
april, line 148: Overall, I don't like that this function returns either a dictionary or `None`. I think it would probably be better if it returned an empty dictionary if there were no results.

(although I don't know why it's a dictionary at all)

fiji-flo: This function return either a dictionary (even an empty one) which reassembles that all access groups should be removed. `None` means this is a no-op.

I could return the unmodified groups and compare in the calling function. I guess this is due to my Rust mental model. If at some point this function would modify the input in some way as a side effect the calling function would not be aware. This reassembles an Option (Rust) / Optional (Java/C++) for me and I think there's no native way in python for that. 

Look at that test cases, all cases should be reflected there.

python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
fiji-flo: Access groups are an object in the profile v2 schema. We never defined how it should look. Right now all access groups are `{ key: null, … }` (ldap/hris/mozilliansorg). To distinguish PMO and Mozillians managed access group during migration the ones managed by PMO will be `{ key: something }` to not override each other and for easy migration.

fiji-flo:
Travis is just flaky: https://travis-ci.com/github/fiji-flo/cis/builds/163025394

PR #500, opened 2020-05-07
fix leeway for token expiration

fiji-flo:


fiji-flo:
worked on my travis …
https://travis-ci.com/github/fiji-flo/cis/builds/164447122

april:
This looks good to me, the only thing I might add is that since `expired` is only used inside this module, maybe the default leeway should be 910 instead of 0?

gene1wood:
Travis builds are likely failing as reported in #492, not because of your changes

fiji-flo:
@april I thought about making the default to 910, but I wanted this explicit in the line its used. But yeah, I get your point.

PR #510, opened 2020-06-17
fix dispaly level filtering

fiji-flo:
- null is not considered public


fiji-flo:
Tests are failing. We need to figure out how this affects behavior.

PR #516, opened 2020-07-17
Clarifying the version of jq required in the PersonAPI docs

gene1wood:


april:
Does it not work for you? I would think that any version of `jq` would work.

april:
Looking at it, I think any version of jq since 2015 (1.5?) should probably work. I'm not really sure that's worth noting.

gene1wood:
Check the link to the release notes in the PR that shows when the `walk` feature was added in 1.6. It's not present prior to 1.6 and so the command doesn't work.

april:
Ahh, you’re right, looks like 1.6. Still, that’s a pretty old release.

gene1wood:
1.5 is what's available to my OS and I encountered errors using the code in the doc and had to troubleshoot the error message about the `walk` feature not working to find changelogs to determine that a specific version of jq is needed, I figured I'd help the next person who encounters the problem by clarifying what version of jq is needed.

april:
Huh, surprising! Sure, looks good to me then. And upgrade your `jq`, it's worth it.  :)

PR #519, opened 2020-09-02
Add metadata service that lets you see if a user is in cis or ldap

april:


gene1wood (commented):
Can you also update the PersonAPI.md to document this new endpoint?

april:
Done!

gene1wood (approved):

PR #521, opened 2020-09-25
make mozilliansorg_groups publisher a noop

fiji-flo:


python-modules/cis_publisher/cis_publisher/mozilliansorg_groups.py
```diff
@@ -1,150 +1,14 @@
...
-        """
-        updated = {group: None for group in mozillians_groups_update}
-        if not current_groups:
-            return updated
-        pmo_groups = {name: meta for name, meta in current_groups.items() if meta is not None}
-
-        updated.update(pmo_groups)
-
-        return updated
+        logger.info("noop")
```
april, line 14: I would maybe have a slightly more descriptive thing here, like “noop on Mozillians group publisher”. But otherwise seems fine.

april (approved):

PR #523, opened 2020-10-29
Better scope identities permissions

april:
Update the identities permissions to allow LDAP to update the LDAP identities.


april:
I would note that the current set of published rules:

https://auth.mozilla.com/.well-known/mozilla-iam-publisher-rules

Allow mozilliansorg to set SSH/PGP keys on creation, but I'm pretty sure that is wrong? Is that indeed wrong, @fiji-flo?

floatingatoll:
Thank you for the time put into this for us, I really appreciate it.

Work continues in #537.

PR #524, opened 2020-12-01
enable deleting profiles on prod

fiji-flo:
Tired of deleting profiles directly in Dynamo.
Require _write_ scope for all change API endpoints.


gcoxmoz:
Quoth gene: "The challenge is that there's no scoping to grant a PersonAPI user rights to do read operations but not deletes. That capability would need to be added first as otherwise all PersonaAPI clients could delete any/all records"

PR #525, opened 2020-12-11
Fix up example curl code to accept variables

gene1wood:


The-smooth-operator (approved):
Looks good, thanks for enhancing the docs

PR #530, opened 2021-07-22
Fix all flake8 errors found in current code

floatingatoll:
Hi, reviewers.

While fixing CI I found a stack of flake8 errors (E###) that were preventing CI from running. I've separated them out for review purposes into this PR, so that they can be checked for sanity.

I have confirmed in #531 that the new CI passes when these are incorporated. It will not pass without these fixes.

### DIFF
--- a/python-modules/cis_identity_vault/integration_tests/test_all_users.py
+++ b/python-modules/cis_identity_vault/integration_tests/test_all_users.py
@@ -45,6 +45,7 @@ def get_all_by_page():
 
     logger.debug("Total records retrieved: {}".format(len(results)))
 
+
 def get_all_by_any():
     setup_environment()
     u = user.Profile(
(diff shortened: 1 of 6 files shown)


floatingatoll:
(This is going to be merged with #529 when I'm done, since otherwise they'll never individually pass CI.)

python-modules/cis_identity_vault/integration_tests/test_all_users.py
```diff
@@ -72,6 +73,7 @@ def get_all_by_any():
         logger.info("Total records retrieved: {}".format(len(results)))
     logger.info("Total records retrieved: {}".format(len(results)))
 
+
 def test_filtered_scan(benchmark):
-    #get_all_by_any()
+    # get_all_by_any()
```
bkochendorfer, line 78: huh, wonder why that's commented out

bkochendorfer (approved):

floatingatoll:
Yeah. I could guess, but it’s outside of scope for CI repair.

duallain (approved):

PR #528, opened 2021-06-10
Add bugzilla.mozilla.org and bugzilla-dev.allizom.org to production notification list

dklawren:

PR #532, opened 2021-07-22
Remove trailing whitespace from a variety of files

floatingatoll:
#528 included fixes for trailing whitespace in a variety of files, which I imagine we should fix regardless of #528 itself. These don't seem to be caught by CI, so this probably will not need to be merged into the CI PR I'm expecting to end up creating out of several other PRs.

Reviewers, my biggest concern is that this _ought_ to be a no-op, but if it will trigger lambda autodeploys or other unknown things, then that would be of concern.


floatingatoll:
Replaced by #540.

PR #531, opened 2021-07-22
Replace travis-ci.org with GitHub Actions

floatingatoll:
The old CI system went away a week or so ago, and per team discussion and in order to unblock #528, we're going to switch this specific repository to GitHub Actions and get CI working again.

Hi, reviewers. This pull request is complicated, as in order to prove that I've repaired the CI system, I also need to incorporate several changes to both the Dockerfile and the CIS code itself.

In this PR, please review _only_ the one unique commit in this PR, 44c79a2, which is not present in any other pull request; it simply removes the Travis CI config and adds the GitHub Actions config. The final push of this pull request will contain only that one unique commit; everything else below is _temporarily_ included from the following pull requests, to demonstrate (as seen below) that CI passes once it's all put together; it's all general repair work that isn't specific to Travis CI or GitHub Actions:

- #534 
- #529 
- #530 

Please do _not_ merge this pull request. I am coordinating several interdependent PRs and will land everything when ready.

PR #534, opened 2021-07-22
Dockerfile repairs and optimizations for CI

floatingatoll:
Hi, reviewers.

These commits were relevant as part of fixing CI for #531, and I've pulled them out separately so that they can be reviewed more easily.

I've already confirmed in #531 that the new CI passes when these changes are incorporated.

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,4 +1,4 @@
-FROM archlinux/base:latest
+FROM archlinux:latest
 LABEL maintainer="akrug@mozilla.com"
 
 RUN pacman -Syyu --noconfirm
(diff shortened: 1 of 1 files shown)


bkochendorfer (approved):

floatingatoll:
I'll proceed with landing these since CI passes with them incorporated in #531 and they're safe to proceed independent of the rest of the work.

floatingatoll:
Note that GitHub managed to wreck the signatures when merging to the repository somehow, even though everything was set up for a zero-modification merge. The repo disallows force pushes, so @gene1wood if you'd like to comment on which repair path I need to take for the commitsigs, I'm certainly interested to hear about it.

PR #533, opened 2021-07-22
Add bugzilla-dev.allizom.org to production notification list (signed commit)

floatingatoll:
This is the signed-commit equivalent of, and, fixes #528.

This contains only the config file change; whitespace fixes are in #532.

Hi, reviewers. Please do _not_ merge this pull request. I am coordinating work across several interdependent PRs.

PR #538, opened 2021-07-27
Add bugzilla-dev.allizom.org to production notification list (signed commit)

floatingatoll:
This is the signed-commit equivalent of, and, fixes #528 (and replaces #533).

This contains only the config file change; whitespace fixes are in #532.

Hi, reviewers. Please do not merge this pull request; we still need to work out the deploy process for CIS first.


bkochendorfer:
This is quite old @floatingatoll. What do you think we should do

PR #541, opened 2021-07-27
ci: Remove -j# from test options

floatingatoll:
We have what looks to be [a CI failure](https://github.com/mozilla-iam/cis/pull/537/checks?check_run_id=3175469630) due to a race condition between two different sets of tests that expect to be using a given port number, so this patch removes all parallelization from the tests (which is how I kept them for most of my repairs anyways). HT @gcoxmoz

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,13 +77,13 @@ push-ci-container:
 .PHONY: test
 test:
 	docker-compose build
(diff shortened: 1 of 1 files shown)


bkochendorfer (approved):
Looks like the tests take about 14minutes now instead of about 9-10 minutes. Not a huge price to pay

bkochendorfer:
I rediscovered this and merged a similar change! I wish I had remembered this yesterday haha

PR #558, opened 2022-06-16
chore(tests): Make them work

bkochendorfer:
There were several issues with these tests preventing them from working.

* The parallelism caused port collision despite using a random port
* The `run_before` section in tox.ini caused several issues in parallel
  as it was killing the testing harness for other tests while running
* There is a bug in flask-restful which prevents it from correctly
  parsing args if you do not specify location and are running inside a
  virtual environment. https://github.com/flask-restful/flask-restful/issues/936
* Moto changed it's testing signature requiring arity of two.
* The psycopg2 manually installed version doesn't seem to work correctly
  in the tests and there are comments saying as much

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,13 +77,13 @@ push-ci-container:
 .PHONY: test
 test:
 	docker-compose build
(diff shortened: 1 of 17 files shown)


floatingatoll (approved):
Yes

PR #559, opened 2022-11-09
Store the HRIS timezone without modification [IAM-989]

floatingatoll:
As noted in IAM-989, we aren't syncing timezones for a lot of workers due to their absence from `tz_convert`'s manual table. IAM-992 removed the only instance of anyone parsing the timezone field, so this updates the HRIS publisher to just copy the Workday-provided timezone into the record without modification. This can be reversed later on if necessary.


gcoxmoz (commented):
All users in the workday report have a Time_Zone (so we're into edge-cases) but, what if they don't.

If `Time_Zone` were ever missing, in the old world the `.get` would ship through `tz_convert` and be collapsed into string `"UTC+0000 Europe/London"`.  Now the timezone.value will be (I presume) pythonic None instead of a string.

You might want to make `.get()` have a UTC-ish default (nitpick, the old default of `Europe/London` isn't wholly UTC because of BST) or an empty string.

floatingatoll:
The CIS profile supports value "null" but I'm fine with empty string.

PR #560, opened 2022-11-17
Fix GitHub CI and tests

floatingatoll:
Hi @bkochendorfer, I'm trying to prepare for a release of CIS for IAM-994, and it took a lot of old-version pinning and various Docker/CI/etc. work to get the tests to mostly pass (except for postgresql, which I guess was already known).

I believe since the work in June, that the deprecation of Python2 on archlinux led to this subsequent "pin old versions now that python3 is in use" effort.

Do you think this is safe to merge and do a serverless deploy from, so that we can start repairing the timezone data?


bkochendorfer (approved):
Good work! Yeah I think you can give it a go if it passses!

PR #562, opened 2024-09-18
IAM-1427, pylint + softly accept miscased user emails in HRIS publisher

gcoxmoz:
2 commits, easier to understand if you examine individually.
commit 1 is pylint inspired cleanup of logging and ordering of imports.
commit 2 is "if you can't match on emails from the workday report, try again with lowercase.  If THAT works, use it but emit a warning; if not, fail like you have been doing."

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,13 +77,13 @@ push-ci-container:
 .PHONY: test
 test:
 	docker-compose build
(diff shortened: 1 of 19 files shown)


gcoxmoz:
Something got mangled, starting over.

PR #575, opened 2025-07-16
We're not returning results fast enough

bheesham:
The HRIS sync job is currently failing because the `v2UsersByAny` endpoint is taking too long to return results. This change bumps the number of segments we use when scanning Dynamo, in an attempt to finish the work _a little bit sooner_.


bheesham:
FYI: @jasonthomas, @Flipez.

I'll merge this and deploy it on dev, and do whatever smoke tests I can there.

PR #576, opened 2025-07-16
deploy Update profile-retrieval to use the new lambda layer, etc

bheesham:


serverless-functions/profile_retrieval/serverless.yml
```diff
@@ -76,9 +79,14 @@ functions:
...
+    # This value is to preserve some manual changes we've had to make along the
+    # way.
+    timeout: 900
+    logRetentionInDays: 30
     events:
       - http: ANY /
       - http: ANY {proxy+}
     layers:
-      -  ${ssm:/iam/cis/${self:custom.profileRetrievalStage}/lambda_layer_arn}
+      -  ${ssm:/iam/cis/${self:custom.profileRetrievalStage}/build/lambda_layer_arn}
```
bheesham, line 92: nitpick: this should include somewhere the architecture and runtime, probably, since we may want to publish multiple lambda layers for different combinations.

Some reasons:

* we may want to test with the same python version but different processor (arm64 should give us some savings);
* we may want to test with a different python but the same architecture (e.g. python upgrades, as we'll soon be forced to do).

I'm not going to address this now, since we're not close to this being an issue. But, it _could_ be an issue.

bheesham:
Deployed using:

```
; SERVERLESS_ACCESS_KEY=... serverless deploy --stage development --region us-west-2 --aws-profile iam-admin
```

bkochendorfer (approved):

bheesham:
I'll roll this out to stage and re-run some tests there. Thankfully I'm able to lean on the tests from https://github.com/mozilla/webservices-infra/tree/main/iam/tf/tests

bheesham:
Following-up: deployed to staging.

PR #579, opened 2025-07-23
Add dax

bheesham:
DAX promises to improve our response times from DynamoDB. This PR makes use of DAX in the route that's currently timing out, causing the HRIS sync job to fail.

A small writeup with results is available on [the internal wiki](https://mozilla-hub.atlassian.net/wiki/spaces/IAM/pages/edit-v2/1704691299). But, the important bits are:

* we don't see an improvement for the failing query on development;
* we may see an improvement on staging, where we have more data.


bheesham:
I'm going to merge this to start work on getting CIS on stage revived. It seems the Dynamo table have the wrong indices configured, so at some point I think the vault curator was ran with incorrect arguments.

PR #581, opened 2025-07-30
Remove DAX from dev and stage

bheesham:
Small writeup: [2025-07-25 DynamoDB Accelerator does not improve response times for Person API](https://mozilla-hub.atlassian.net/wiki/spaces/IAM/pages/1704691299/2025-07-25+DynamoDB+Accelerator+does+not+improve+response+times+for+Person+API)

We found that adding DAX does not improve response times after trying a couple of different things. The writeup has more deets, but the gist of it is:

* we can see profile retrieval making use of it;
* we can see we're hitting the cache, and not going to Dynamo;
* we can see response times worsen.

Same behaviour when trying with different replicas.

I'm including the code here for posterity, should we want to revive this at some point. Plus, there were a couple of bug fixes along the way which seemed like a good idea to include.

Jira: [IAM-1732](https://mozilla-hub.atlassian.net/browse/IAM-1732)


bkochendorfer (approved):
Thank you for cleaning up the experiment

PR #583, opened 2025-08-01
Prep profile-retrieval for containerization

bheesham:
See commit(s).

---

Jira: [IAM-1735](https://mozilla-hub.atlassian.net/browse/IAM-1735)


skaffold.yaml
```diff
@@ -0,0 +1,53 @@
...
+      hooks:
+        # Build what we test. We could opt for test what we build as well.
+        before:
+          - command: ["./ci/test.sh"]
+            os: ["linux", "darwin"]
+deploy:
+  kubeContext: docker-desktop
+  helm:
+    releases:
+      - name: profile-retrieval-api
```
bkochendorfer, line 44: This is interesting. We normally put the Helm charts for deploying in the *services-infra repo. If you are okay with having a smaller version here in the repository for local testing I think that is :ok:

bkochendorfer (approved):

skaffold.yaml
bheesham: Yeah, this is mostly for local testing.

My main goal here is to keep _a_ Helm chart close to where I'm going to be developing the code. It'll require some copy/pasting, but keeps me from having to context switch until I need to.

PR #594, opened 2025-09-16
Fix a bug with pagination in Dynamo's parallel scan

bheesham:
As it turns out, `cis_profile_retrieval_service` was doing the right thing. But, `cis_identity_vault` was not.

Three main issues solved in this PR:

1. `cis_identity_vault`'s parallel scan was paginating for callers, breaking the expectation that pagination was being used, or could have been used.
2. The last page of parallel scans _may_ include duplicate results. At least in the test using mocks. It's unclear if this is the case when running against a real DynamoDB.
3. Adding tests again for `cis_profile_retrieval_service`, which atrophied.

The commit messages are more descriptive than this PR.

Jira: [IAM-1793](https://mozilla-hub.atlassian.net/browse/IAM-1793)

### DIFF
--- a/python-modules/cis_identity_vault/cis_identity_vault/models/user.py
+++ b/python-modules/cis_identity_vault/cis_identity_vault/models/user.py
@@ -317,15 +317,65 @@ class Profile(object):
             users.extend(response["Items"])
         return users
 
-    def _last_evaluated_to_friendly(self, last_evaluated_key):
-        if last_evaluated_key is None:
+    def _last_evaluated_to_friendly(self, last_evaluated_keys):
+        """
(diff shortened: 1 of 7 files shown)


bheesham:
I wrote an integration test, using local code and staging, which revealed some extra weirdness around segments not having the correct ESK.

I've got a fix locally, though I need to clean it up a bit.

bheesham:
Rebased and pushed! The current branch _should_ fix pagination with multiple segments.

Flipez (approved):

bheesham:
@Flipez and I did a demo and review over Slack. I'll spruce up some of the comments, because this was hard to figure out the first time and I'd rather us not pay that cost again.

PR #596, opened 2025-09-23
fix(publishers): code accesses status_code; ok is a property not a fn

bheesham:
Jira: Untracked

---

See: https://github.com/mozilla-iam/cis/actions/runs/17948809950/job/51042131282#step:6:457


gcoxmoz (approved):
Makes sense, and is 'just' tests, I'm happy if you are.

PR #597, opened 2025-09-23
fix(identity-vault): return nextPage tokens which are API Gateway friendly

bheesham:
> The plain text pipe character (|) and the curly brace character ({ or }) are
> not supported for any request URL query string and must be URL-encoded.

via [Amazon API Gateway important notes][important-notes]

Jira: [IAM-1793](https://mozilla-hub.atlassian.net/browse/IAM-1793)

[important-notes]: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html#api-gateway-known-issues-rest-apis


python-modules/cis_identity_vault/cis_identity_vault/models/user.py
```diff
@@ -348,11 +349,12 @@ def _last_evaluated_to_friendly(self, last_evaluated_keys):
...
         # If there are at all any segments left with work, then continue.
         if any(next_page):
-            return ",".join(next_page)
+            next_page_raw = ",".join(next_page)
+            return urllib.parse.quote(next_page_raw)
         else:
             return None
 
-    def _next_page_to_dynamodb(self, next_page):
+    def _next_page_to_dynamodb(self, next_page_raw):
```
gcoxmoz: The most minor of quibbles here that you're passing by position, a caller that's `next_page`, into a parameter `next_page_raw`, but I can't hold you up on that.

gcoxmoz (approved):
Makes sense.
Does a unit test make sense to add here?

bheesham:
> Does a unit test make sense to add here?

Added!

PR #598, opened 2025-10-14
fix(hris): staff aren't being updated often enough

bheesham:
The job doesn't run often enough, leaving staff sometimes up to 6 hours out of date.

There were once rumours of rate limits around HRIS's Reports As A Service (RAAS) API, and so if we'd like to reduce this further we'll need to sync with the people team.

Jira: [IAM-1821](https://mozilla-hub.atlassian.net/browse/IAM-1821)


gcoxmoz (approved):
Tracks with previous discussions / attempts to improve day-1 experiences.

PR #611, opened 2026-08-05
chore: use the WAF URLs

bheesham:
We need these systems to agree on an issuer.

Also will need to update the LDAP publisher, which is in a different repo.

Jira: IAM-2055


gcoxmoz (approved):
Prod isn't touched and you've been doing a bunch of these already today so 'have fun!'

